Issue #10, opened 2019-02-25
An error occurs when sending messages into the Palo Alto input


danotorrey:
An error occurs when sending messages into the Palo Alto input. The specific error is unknown, but we are working to obtain it. A PCAP has also been obtained, which is being used to investigate the issue further. 

More details will be added to this issue shortly as the specific issue that is occurring is identified.

danotorrey:
I am not able to reproduce any error when replaying the following message to the Palo Alto input:

@waynekearns Can we please collect a debug log from when the issue happens? Here are [instructions for enabling debug logging for a plugin](http://docs.graylog.org/en/3.0/pages/faq.html#how-do-i-enable-debug-logging-for-a-specific-plugin-or-area-of-graylog). Here's the CURL command for reference:

```
curl -I -X PUT http://username:password@localhost:9000/api/system/loggers/org.graylog.integrations/level/debug \
-H 'X-Requested-By: graylog-api-user' \
-X PUT \
-I
```

This should help provide clarity on the specific issue that is occurring.

danotorrey:
We are making progress on this issue, but it is not solved yet. The first major issue was the the Palo Alto input was using the TcpTransport instead of the correct SyslogTcpTransport. This fixed part of the issue, but there is still a parsing error that only occurs with actual packets in prod and not the text version of the message exported from Wireshark. I’m planning to replay the PCAP first thing in the morning to identify and fix the remaining issue.

danotorrey:
I have reproduced the issue with PCAP replay and am working on a fix.

danotorrey:
@waynekearns  has tested and confirmed that the Palo Alto bug is fixed. 👍

I’ll submit a PR with tests today, and we’ll include it in the next bug fix release. It is OK to continue using the snapshot JAR until then.

bernd:
Fixed in #11

PR #11, opened 2019-02-25
Fix Palo Alto input parsing issue

danotorrey:
Fixes issue #10.


danotorrey:
@bernd @deeshe My feeling is that it is ok to include this fix in release 3.0.1 and not to backport it. What do you think? It probably would not hurt to backport, but not sure if it is necessary.

deeshe:
@danotorrey The fix should definitely be included in 3.0.1 as @kroepke said. 
The project and labels are perfect, thank you! 👍

danotorrey:
This change will be included in 3.0.1 and backported to 2.5.

PR #80, opened 2019-06-13
Add Available Services API call

### LINKED ISSUE #50: Add API call for list of available AWS Services
The implementation of this might require:

- [ ] Resource method with annotations.
- [ ] Service method in AWSService. Would probably return List<AWSService> 

AWSService class might have these properties:

String name
String description
String policyPermission
String helperText
String learnMoreLink

To support this page:

image

danotorrey:
Add API call that returns a list of available services for the Choose from Available Services page. 

To test this out, this CURL command can be used: `curl http://someuser:somepass@localhost:9000/api/plugins/org.graylog.integrations/aws/availableServices`

Resolves #50


danotorrey:
The linter check failure can be ignored, since this code did not change any UI components.

danotorrey:
Due to some incorrect branching, closing this PR and cherry-picking into another PR.

PR #84, opened 2019-06-18
Shared State w/ Hooks

kyleknighted:
Bypassed the reliance of Reflux to just use Hooks to manage the state of the fields.


src/web/aws-cloudwatch/CloudWatchContext.jsx
```diff
@@ -0,0 +1,114 @@
...
+          ...state,
+          availableSteps: action.value,
+        };
+      }
+
+      default: return state;
+    }
+  };
+
+  const [state, dispatch] = useReducer(reducer, initialState);
```
dennisoelkers: This reducer is managing three things: 1) wizard steps, 2) form data, 3) advanced options visibility. All three are not logically coupled, but they are now structurally. I think it would be better at this point to separate all three into simpler, independent state-keepers (probably using `useState` instead, at least for current/available/enabled steps & the advanced options toggle and structure it into separate contexts, specific to each concern.

src/web/aws-cloudwatch/FormAdvancedOptions.jsx
```diff
@@ -14,32 +18,32 @@ const FormAdvancedOptions = ({ onChange, values, toggle, visible }) => {
         <Input id="awsCloudWatchGlobalInput"
                type="checkbox"
                value="global-input"
-               defaultChecked={values.awsCloudWatchGlobalInput}
+               defaultChecked={getFieldValue('awsCloudWatchGlobalInput')}
```
dennisoelkers: Is it possible to turn this into a object returned by `formDataHook` that we can destructure, so we do not have to do a function call for every input?

kyleknighted: Hopefully you have a better idea for me, because this was the cleanest I could make it.
If I replace the functions, then every value ends up being `formData.awsCloudWatchName.value || formData.awsCloudWatchName.defaultValue`

src/web/aws-cloudwatch/reducers/StepsContext.jsx
```diff
@@ -0,0 +1,60 @@
...
+          ...state,
+          availableSteps: action.value,
+        };
+      }
+
+      default: return state;
+    }
+  };
+
+  const [state, dispatch] = useReducer(reducer, initialState);
```
dennisoelkers: In this case we can also decouple those three, independent variables into separate states like this:

```
const [currentStep, setCurrentStep] = useState('authorize');
const [enabledSteps, enableStep] = useState(['authorize']);
const [availableSteps, makeStepsAvailable] = useState([]);

[...]
<StepsContext.Provider value={{ currentStep, setCurrentStep, enabledSteps, enableStep, availableSteps, makeStepsAvailable }}>
```

This way those variables are handled independently and we have already have those rich, abstracted (in contrast to `state`/`dispatch`) setters that are now created in `hooks/step.js`. We could even separate the `StepsContext` into three smaller contexts if we want to.

Even if we decide that those variables need to be coupled at some point, we can still do it by creating and exporting a different setter that cascades updates, or even switch back to a reducer again and create and export the same setters again.

src/web/aws-cloudwatch/CloudWatch.jsx
```diff
@@ -1,169 +1,101 @@
...
-    return !enabledSteps.includes(step);
+  const handleEditClick = nextStep => () => {
+    setCurrentStep(nextStep);
   };
 
-  handleFieldUpdate = ({ target }) => {
-    const { formData } = this.state;
+  const handleFieldUpdate = ({ target }) => {
+    const id = target.name || target.id;
     const value = Object.keys(target).includes('checked') ? target.checked : target.value;
```
dennisoelkers: You can also use `FormUtils.getValueFromInput` to retrieve an input's value based on its type.

src/web/aws-cloudwatch/context/default_settings.js
```diff
@@ -0,0 +1,37 @@
...
+    value: 'Name',
+  },
+  awsCloudWatchDescription: {
+    value: 'Description',
+  },
+  awsCloudWatchAwsKey: {
+    value: 'AKQQQQQQQQQQQQQQQQQQ',
+  },
+  awsCloudWatchAwsSecret: {
+    value: 'XfoodLP3YXdzX3NlY3JldF8wMDNOTDAwMDbeMA==',
```
dennisoelkers: Maybe next time you can put credentials in a separate file you are including for development, which is not checked in into github?

src/web/aws-cloudwatch/StepHealthCheck.jsx
```diff
@@ -1,14 +1,19 @@
...
+const StepHealthCheck = ({ onSubmit }) => {
+  const { logOutput } = useContext(LogOutputContext);
+
   return (
     <Row>
       <Col md={8}>
-        <form onSubmit={onSubmit}>
+        <FormWrap onSubmit={onSubmit} buttonContent="Review &amp; Finalize">
           <h2>Create Kinesis Stream</h2>
           <p>We&apos;re going to attempt to parse a single log to help you out! If we&apos;re unable to, or you would like it parsed differently, head on over to <a href="/system/pipelines">Pipeline Rules</a> to set up your own parser!</p>
```
dennisoelkers: One thing I noticed: We should either open the link in a new tab or warn the user about losing the values entered in the form.

src/web/aws-cloudwatch/StepReview.jsx
```diff
@@ -1,68 +1,110 @@
...
-              <li><strong>Stream</strong><span>{values.awsCloudWatchKinesisStream}</span></li>
-              <li><strong>Global Input</strong><span>{values.awsCloudWatchGlobalInput ? 'true' : 'false'}</span></li>
-              <li><strong>AWS Assumed ARN Role</strong><span>{values.awsCloudWatchAssumeARN || 'None'}</span></li>
+              <li>
+                <strong>Stream</strong>
+                <span>{formData.awsCloudWatchKinesisStream.value}</span>
+              </li>
+              <li>
+                <strong>Global Input</strong>
+                <span>{(formData.awsCloudWatchGlobalInput && formData.awsCloudWatchGlobalInput.value) ? 'true' : 'false'}</span>
```
dennisoelkers: We could also use a glyph (checkmark/cross) in here?

src/web/aws-cloudwatch/StepReview.jsx
```diff
@@ -1,68 +1,110 @@
...
-  const defaultOutput = (key, enabled = true) => {
-    if (!enabled) {
-      return (
-        <React.Fragment>
-          {DEFAULT_VALUES[key]} <small>(default)</small>
-        </React.Fragment>
-      );
-    }
+const format = (numString) => {
+  return Number(numString).toLocaleString('en-US');
```
dennisoelkers: Maybe we should leave out the locales so we are defaulting to the browser/os setting?

src/web/aws-cloudwatch/StepHealthCheck.jsx
kyleknighted: This will be handled in https://github.com/Graylog2/graylog-plugin-integrations/issues/107

PR #85, opened 2019-06-18
Form Field Validation

kyleknighted:
Provides help message and validation to each field.

Could add a list of errors at the top of the form as well, but this should work for the time being.
Feel free to share anything that may already exist for field validation.


src/web/aws-cloudwatch/CloudWatch.jsx
```diff
@@ -2,12 +2,12 @@ import React, { useContext } from 'react';
 
 import Wizard from 'components/common/Wizard';
 import FormUtils from 'util/FormsUtils.js';
+import formValidation from 'utils/formValidation';
```
kyleknighted: Maybe this stuff moves to core `FormUtils`?

src/web/aws-cloudwatch/StepAuthorize.jsx
```diff
@@ -23,57 +24,60 @@ const StepAuthorize = ({ onChange, onSubmit }) => {
...
-                 label="AWS Key"
-                 placeholder="CloudWatch Integration AWS Key"
-                 onChange={onChange}
-                 value={formData.awsCloudWatchAwsKey ? formData.awsCloudWatchAwsKey.value : ''}
-                 autoComplete="off"
-                 pattern="AK[A-Z0-9]{18}"
-                 minLength="20"
-                 required />
+          <ValidatedInput id="awsCloudWatchAwsKey"
+                          type="password"
```
danotorrey: I believe that the AWS Key field should be unmasked. This is how it shows in the AWS console. This helps the user to confirm they are pasting the correct credential set. The AWS Secret should continue to be masked.

src/web/utils/formValidation.js
```diff
@@ -1,4 +1,35 @@
+import _ from 'lodash';
+
 const formValidation = {
+  checkInputValidity: (input, customErrorMessage) => {
```
danotorrey: It looks like the value for text fields is cleared when when the validation is triggered. See example gif below. Is this intentional? I think we should consider keeping the value present, so the user can inspect what they entered to confirm the issue (eg. check if they only partially copied a key or secret).

Jul-10-2019 17-01-13

src/web/aws-cloudwatch/StepAuthorize.jsx
```diff
@@ -23,57 +24,60 @@ const StepAuthorize = ({ onChange, onSubmit }) => {
...
-                 type="password"
-                 label="AWS Key"
-                 placeholder="CloudWatch Integration AWS Key"
-                 onChange={onChange}
-                 value={formData.awsCloudWatchAwsKey ? formData.awsCloudWatchAwsKey.value : ''}
-                 autoComplete="off"
-                 pattern="AK[A-Z0-9]{18}"
-                 minLength="20"
-                 required />
+          <ValidatedInput id="awsCloudWatchAwsKey"
```
danotorrey: I think the validations for the AWS Key and AWS Secret here are too specific (checking that exactly 20/40 characters in length and a key starting with "AK"). I definitely see that from several key/secret pair examples, this seems to be true. However, the [official documentation (Access Keys section)](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html) does not indicate that this will always be the case. What if a key or secret is generated in the future that does not exactly match these checks? I just want to make sure we don't have a blocker issue come in at a later date due to a user not being able to enter their key/secret.

I think we should only check whether they have entered a non-blank value for the Key/Secret. The helper text below the field indicating the length/pattern expectations should be enough to tell the user what they're looking to enter. We'll fully check the key/secret anyhow when performing other API calls. 

Looping in @ceruleancee for input and reference too.

src/web/aws-cloudwatch/StepAuthorize.jsx
kyleknighted: > I believe that the AWS Key field should be unmasked.

Should we unmask the Review page as well?

src/web/aws-cloudwatch/StepAuthorize.jsx
kyleknighted: _Carry over from PMs for posterity._

Claudia and I discussed the Key and Secret patterns and went with a simplified version of: https://aws.amazon.com/blogs/security/a-safer-way-to-distribute-aws-credentials-to-ec2/

src/web/utils/formValidation.js
kyleknighted: _Carry over from PMs for posterity._

This rendered a centered form makes me think there is a caching issue locally for you, because these styles were removed. I am also unable to replicate the disappearing fields in the latest Firefox and Chrome.

PR #93, opened 2019-06-21
Save AWS input API endpoint

### LINKED ISSUE #51: Add API call to save final CloudWatch configuration
When clicking the Complete CloudWatch Setup button, this API call should save everything as an input.

Also include the ability to edit the saved input.

image

danotorrey:
Adds the ability to save an AWS input after the setup process is complete.

Lots of refactoring in healthCheck and message detection was performed to allow for an input to be saved with the appropriate type of message identified. 

Please test the healthCheck call once again when reviewing this PR.

Resolves #51

Note that the layout and structure of the AWSInput and AWSMetaCodec (which will eventually pick the appropriate kind of codec based on the AWSMessageType) needs more work. This is M2 activity, and I would like to work more on this in a subsequent PR (this one is big enough as it is 😆).


danotorrey:
From looking at the file diffs, it looks like there might still be some auto-formatting differences between my Intellij and @ceruleancee's 😃 We'll take a look at this.

src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java
```diff
@@ -0,0 +1,167 @@
...
+import software.amazon.awssdk.regions.Region;
+
+import javax.inject.Inject;
+
+/**
+ * General AWS input for all types of supported AWS logs.
+ */
+public class AWSInput extends MessageInput {
+
+    public static final String NAME = "AWS Input";
```
bernd: I think naming it `AWS` or `AWS Logs` should be enough. :smile: Adding an "Input" suffix seems redundant to me.

src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java
```diff
@@ -0,0 +1,167 @@
...
+              codec.create(configuration),
+              config,
+              descriptor,
+              serverStatus);
+    }
+
+    @Override
+    public void launch(InputBuffer buffer) throws MisfireException {
+
+        LOG.info("Starting AWS Input...");
```
bernd: We already have logging statements for starting inputs so I think this can go because it doesn't provide any further insights.

src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java
```diff
@@ -0,0 +1,167 @@
...
+        super.launch(buffer);
+    }
+
+    @Override
+    public void stop() {
+
+        LOG.info("Stopping AWS Input...");
```
bernd: We already have logging statements for stopping inputs so I think this can go because it doesn't provide any further insights.

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
```diff
@@ -118,4 +124,43 @@ public Response kinesisHealthCheck(@ApiParam(name = "JSON body", required = true
...
+     *      "stream_name": "flow-logs",
+     *      "batch_size": 10000,
+     *      "assume_role_arn": "",
+     *      "global": false,
+     *      "enable_throttling": false
+     * }'
+     */
+    @POST
+    @Timed
+    @Path("/save")
```
bernd: How about using `/aws/inputs` as path for saving inputs? That's a bit more specific than `/aws/save`.

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
```diff
@@ -118,4 +124,43 @@ public Response kinesisHealthCheck(@ApiParam(name = "JSON body", required = true
...
+    @ApiOperation( value = "Save a new Kinesis input" )
```
bernd: Is this a Kinesis specific resource? If so we should probably change the HTTP path to be more specific. If not, we should probably change the description here.

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
```diff
@@ -118,4 +124,43 @@ public Response kinesisHealthCheck(@ApiParam(name = "JSON body", required = true
...
+    @RequiresPermissions(RestPermissions.INPUTS_CREATE)
+    @AuditEvent(type = AuditEventTypes.MESSAGE_INPUT_CREATE)
+    public Response create(@ApiParam(name = "JSON body", required = true)
+                           @Valid @NotNull AWSInputCreateRequest saveRequest) throws Exception {
+
+        awsService.saveInput(saveRequest, getCurrentUser());
+
+        // TODO: Identify if this method needs to return a specific response with the id of the new input.
+        return Response.ok().build();
```
bernd: Our regular input resource only returns the newly created object ID but in other resources we return the newly created entity. I think the latter makes sense here. That way the caller gets access to the new entity without having to do another GET request to get the details.

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
```diff
@@ -351,26 +359,26 @@ private KinesisHealthCheckResponse detectAndParseMessage(String logMessage, Stri
...
 
-        return KinesisHealthCheckResponse.create(true, awsLogMessage.detectLogMessageType().toString(),
+        return KinesisHealthCheckResponse.create(true, awsMessageType,
                                                  responseMessage,
                                                  buildMessageSummary(fullyParsedMessage, logEvent.message()));
     }
 
     /**
-     * Prepare a string summary of all fields. This will be displayed on the Health Check results page.
+     * Prepare a string summary of all fields. This wil be displayed on the Health Check results page.
```
bernd: Typo :smile:

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
danotorrey: Makes sense. For now, I have changed this to reflect specifically a Kinesis AWS input creation. Each future AWS input type will likely have its own collection of unique required fields. I think separate request value objects should be maintained for each input type (to avoid one massive/confusing common object). 

That being said, there are two types of Kinesis inputs that I want to discuss to sort out a few details with them. I'll message you when I get online tomorrow.

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
danotorrey: I do not see the value in returning the entire entity in this case. All of the relevant input details were just posted to this endpoint, so the caller still has those available. Returning the ID definitely make sense though. 

If we were to return the input entity, I think we would need to convert it to an InputSummary value object [similarly to as is being done in InputsResource](https://github.com/Graylog2/graylog2-server/blob/a9a1df0ec08f1b82c973a46446adecd7fd6c9732/graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/InputsResource.java#L207). I am just wondering if it's really worth duplicating this block of code right now. What do you think? Definitely open to feedback on this.

src/main/java/org/graylog/integrations/aws/AWSMessageType.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public boolean isKinesis() {
+        return this.source.equals(Source.KINESIS);
+    }
+
+    public String getLabel() {
+        return label;
+    }
+
+    public boolean isFlowLog() {
```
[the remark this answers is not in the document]
danotorrey: 100% agree.

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
bernd: > I do not see the value in returning the entire entity in this case. All of the relevant input details were just posted to this endpoint, so the caller still has those available. Returning the ID definitely make sense though.

This is more about API conventions than solving an issue for this specific use case. In HTTP APIs you usually get the newly created entity back when you create it. That's what we do in most places of our API. (the inputs resource being one exception)

I leave it up to you. :smiley: 

> If we were to return the input entity, I think we would need to convert it to an InputSummary value object [similarly to as is being done in InputsResource](https://github.com/Graylog2/graylog2-server/blob/a9a1df0ec08f1b82c973a46446adecd7fd6c9732/graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/InputsResource.java#L207). I am just wondering if it's really worth duplicating this block of code right now. What do you think? Definitely open to feedback on this.

If we return the newly created entity, we have to duplicate this piece of code or update the server code to make it shareable. The code is masking passwords to avoid exposing them via the API.

bernd:
Can you please fix the conflicts? :smile:

danotorrey:
> Can you please fix the conflicts? 😄

Will do :)

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
danotorrey: 100% understood. Thanks for the details. This makes sense. I agree with making this consistent with our other API calls and with APIs in general. I'll add the code to return the entity and will follow this pattern in the future too.

danotorrey:
@bernd The merge conflicts have been resolved.

danotorrey:
In order for the checks on this PR to pass, server PR https://github.com/Graylog2/graylog2-server/pull/6047 must be merged.

src/main/java/org/graylog/integrations/aws/resources/AWSResource.java
danotorrey: This has been addressed with https://github.com/Graylog2/graylog-plugin-integrations/pull/93/commits/1fc1735b00a2707331cc71542fbc3292f1d1c695 and server PR https://github.com/Graylog2/graylog2-server/pull/6047.

src/main/java/org/graylog/integrations/aws/codecs/AWSMetaCodec.java
```diff
@@ -0,0 +1,59 @@
...
+import org.graylog2.plugin.inputs.annotations.ConfigClass;
+import org.graylog2.plugin.inputs.annotations.FactoryClass;
+import org.graylog2.plugin.inputs.codecs.AbstractCodec;
+import org.graylog2.plugin.inputs.codecs.Codec;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+
+public class AWSMetaCodec extends KinesisLogDataCodec {
```
bernd: Should we name this `AWSKinesisMetaCodec`? It extends `KinesisLogDataCodec` so it's not really a generic AWS meta codec, right?

src/main/java/org/graylog/integrations/aws/codecs/AWSMetaCodec.java
```diff
@@ -0,0 +1,59 @@
...
+    public static final String NAME = "AWSMetaCodec";
```
bernd: Same as above.

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
```diff
@@ -138,7 +136,7 @@ public HealthCheckResponse healthCheck(KinesisHealthCheckRequest request) throws
         }
 
         // Detect the type of message
-        return detectAndParseMessage(new String(payloadBytes), request.streamName(), request.logGroupName());
+        return detectAndParseMessage(new String(payloadBytes), DateTime.now().getMillis() / 1000, request.streamName(), "", "");
```
bernd: What's the reason for using a `long` here instead of a `DateTime`? Also, why only use second precision instead of millisecond precision?

I would suggest to pass a `DateTime.now(DateTimeZone.UTC)` to `detectAndParseMessage()` and change the `timestamp()` field in `KinesisLogEntry` to `DateTime`. That way we don't have to convert the `long` back to a `DateTime` later and we don't lose any precision.

src/main/java/org/graylog/integrations/aws/codecs/AWSMetaCodec.java
danotorrey: Thanks for noticing this. I've removed the `KinesisLogDataCodec` extension. This was actually intended to be be a general AWS codec since there will be one AWSInput.

`KinesisLogDataCodec` has been renamed to `AbstractKinesisCodec` (the name had always confused me, even in the older AWS plugin). This class is only used for the two actual Kinesis codecs `KinesisRawLogCodec` and `KinesisCloudWatchFlowLogCodec`.

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
danotorrey: Agreed and fixed. I also discovered that the Kinesis record has the approximate arrival time of the message in the Kinesis stream. I've changed the raw Kinesis message detection to use this instead. I figured this was better than using the time when Graylog read the message (even though it only has ms precision). Ideally, the user could provide their own timestamp with a raw message, but there's no easy solution for this.

https://github.com/Graylog2/graylog-plugin-integrations/blob/e7f0fa3fb75aea9a5b82bb72c04a7cdc26c4aaab/src/main/java/org/graylog/integrations/aws/service/KinesisService.java#L140-L143

PR #112, opened 2019-06-25
Allow optional Log Group and Stream

danotorrey:
CloudWatch Log Group and Log Stream are optional, since messages may have been written directly to Kinesis without using CloudWatch. Only CloudWatch messages written VIA Kinesis CloudWatch subscriptions will contain a log group and stream.

`Optional.ifPresent` is used since field assignment is still conditionally required see:
https://github.com/Graylog2/graylog-plugin-integrations/blob/21c41377090a36907510d78f4bc87387db396377/src/main/java/org/graylog/integrations/aws/codec/CloudWatchFlowLogCodec.java#L53).


danotorrey:
I've hit some issues with the use of Optional for value object and will continue to investigate tomorrow.

PR #113, opened 2019-06-26
Aws permissions check

### LINKED ISSUE #88: Add permissions checking for AWS API calls
Add permissions checking for AWS API calls.

ceruleancee:
closes #88

### DIFF
--- a/src/main/java/org/graylog/integrations/aws/AWSPermissions.java
+++ b/src/main/java/org/graylog/integrations/aws/AWSPermissions.java
@@ -1,0 +1,38 @@
+package org.graylog.integrations.aws;
+
+import com.google.common.collect.ImmutableSet;
+import org.graylog2.plugin.security.Permission;
+import org.graylog2.plugin.security.PluginPermissions;
+
+import java.util.Collections;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


src/main/java/org/graylog/integrations/aws/AWSPermissions.java
```diff
@@ -0,0 +1,36 @@
...
+import org.graylog2.plugin.security.Permission;
+import org.graylog2.plugin.security.PluginPermissions;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.graylog2.plugin.security.Permission.create;
+
+public class AWSPermissions implements PluginPermissions {
```
bernd: You have to register the custom permissions in the `IntegrationsModule` with `addPermissions(AWSPermissions.class)`. Otherwise they will not show up in the REST endpoints that return all available permissions.

danotorrey: I know that @ceruleancee had done this. Perhaps the file did not get committed. I committed this change now, so this PR can be merged. 🙂

PR #114, opened 2019-06-27
Return JSON message field list in response from Health Check

### LINKED ISSUE #109: Return message JSON object for Health Check
Health Check is currently returning a `message_summary` field which is a string list of fields separated by line breaks. This needs to be changed to a JSON object to allow for improved display in the UI.

danotorrey:
This changes the Health Check response to return a list of message fields instead of the text summary. Resolves #109

New example response:
```
{
  "success": true,
  "input_type": "KINESIS_FLOW_LOGS",
  "explanation": "Success. The message is a Flow Log message.",
  "message_fields": {
    "protocol_number": 6,
    "src_addr": "172.1.1.2",
    "source": "aws-flowlogs",
    "message": "eni-3244234 ACCEPT TCP 172.1.1.2:80 -> 172.1.1.2:2264",
    "dst_addr": "172.1.1.2",
    "dst_port": 2264,
    "action": "ACCEPT",
    "_id": "4ecfcb60-9913-11e9-a3a6-32ca2bdd4771",
    "aws_log_stream": "eni-3423-all",
    "timestamp": "2019-06-05T12:35:44.000Z"
  }
}
```

The plan is to display these fields in a formatted way in the UI instead of in a big block of text.

### DIFF
--- a/src/main/java/org/graylog/integrations/aws/AWSMessageType.java
+++ b/src/main/java/org/graylog/integrations/aws/AWSMessageType.java
@@ -1,5 +1,6 @@
 package org.graylog.integrations.aws;
 
+import org.graylog.integrations.aws.codecs.KinesisCloudWatchFlowLogCodec;
 import org.graylog.integrations.aws.codecs.KinesisRawLogCodec;
 import org.graylog.integrations.aws.transports.KinesisTransport;
 import org.graylog2.plugin.inputs.codecs.AbstractCodec;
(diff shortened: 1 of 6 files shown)


src/main/java/org/graylog/integrations/aws/AWSMessageType.java
```diff
@@ -21,8 +22,8 @@
                 AbstractCodec.Factory.class, KinesisTransport.Factory.class),
 
     // Flow Logs delivered to Kinesis via CloudWatch subscriptions.
-    KINESIS_FLOW_LOGS(Source.KINESIS, "Flow Log", KinesisRawLogCodec.NAME,
-                      KinesisRawLogCodec.Factory.class, KinesisTransport.Factory.class),
+    KINESIS_FLOW_LOGS(Source.KINESIS, "Flow Log", KinesisCloudWatchFlowLogCodec.NAME,
```
danotorrey: The incorrect Flow Log codec was specified here. Fixed and tests added.

kyleknighted:
I can't comment on the code itself, but I can say that this output is much more preferred than raw text. 👍

danotorrey:
> I can't comment on the code itself, but I can say that this output is much more preferred than raw text. 👍

Excellent! Thanks @kyleknighted!

Issue #27, opened 2019-04-30
Create general AWS input


danotorrey:
Create a general input for AWS (named "General AWS"). 

In the initial version, the user will choose this General AWS input from the list of available inputs on the Inputs page. We will want it to be clear that this is different from the existing AWS plugin inputs listed [here](https://github.com/Graylog2/graylog-plugin-aws).

Some of this input code will be written in #51.

This should tie together the following issues:

* Kinesis Transport https://github.com/Graylog2/graylog-plugin-integrations/issues/71
* Kinesis Codec https://github.com/Graylog2/graylog-plugin-integrations/issues/70

danotorrey:
Note that three separate AWS SDK permissions are needed: logs:DescribeLogGroups, logs:DescribeLogSteams, logs:GetLogEvents. The describe actions are only to display a list of available choices in the UI. If these permissions are not available, we should ask the user to manually input a value/values (instead of displaying an error). logs:GetLogEvents will ultimately be used later in the flow to perform a test log retrieval to ensure everything works end to end.

danotorrey:
I think we can break this issue up in to smaller pieces.

PR #116, opened 2019-06-28
Implement AWS codec selector with tests

### LINKED ISSUE #70: Create Kinesis codec
Create the codec that will decode AWS CloudWatch log messages. This will need to look at the input settings to identify which log format was identified during the setup process.

danotorrey:
This adds the logic, which chooses the correct codec for a particular [AWSMessageType](https://github.com/Graylog2/graylog-plugin-integrations/blob/494982d3bbab7778ea52a694b7f50d75f7fdbdfc/src/main/java/org/graylog/integrations/aws/AWSMessageType.java#L21-L26).

Closes #70 #27


src/main/java/org/graylog/integrations/aws/codecs/AWSCodec.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import java.util.Map;
+
+public class AWSCodec extends AbstractCodec {
+
+    public static final String NAME = "AWSMetaCodec";
```
bernd: How about renaming this to `AWSCodec` too?

```suggestion
    public static final String NAME = "AWSCodec";
```

src/test/java/org.graylog.integrations/AWSTestingUtils.java
```diff
@@ -0,0 +1,60 @@
+package org.graylog.integrations;
```
bernd: It looks like this file resides in the `src/test/java/org.graylog.integrations/` directory. Usually this is a nested directory structure. (e.g. `src/test/javaorg/graylog/integrations/`)

(same for all other test files in this PR)

Is this on purpose?

danotorrey: Thanks for catching this. Fixed in https://github.com/Graylog2/graylog-plugin-integrations/pull/116/commits/1d8c3156128f30ccca992a85a1f8fe7b768c8ae6

PR #132, opened 2019-07-11
Fetch data from API

kyleknighted:
- Introduces the API calls for Regions, Streams, and Health Check.
- Has very generic loading state to disable buttons
- Creates custom hook for generic fetch handling with hook support
- Introduces an mechanic to have AWS keys locally without worrying about copy/paste or putting them in a PR.
- Does _not_ save yet, that will be in a followup PR


danotorrey:
Wahoo! I'll review first thing on Monday morn.

danotorrey:
@kyleknighted I am getting this error when trying to click the Authorize & Choose Stream button on the `/integrations/aws/cloudwatch` page.

image

Looks like the error is happening when trying to post to `/api/plugins/org.graylog.integrations/aws/kinesis/streams` with the following body. I think the value/dirty sub fields might be causing the issue here.

```
{
  "aws_access_key_id": {
    "value": "key",
    "dirty": true
  },
  "aws_secret_access_key": {
    "value": "secret",
    "dirty": true
  },
  "region": "eu-west-1"
}
```

src/web/aws-cloudwatch/hooks/fetch.js
```diff
@@ -0,0 +1,81 @@
...
+
+    return () => {
+      didCancel = true;
+    };
+  }, [qualifiedURL]);
+
+  return [{ loading, error, data }, setFetchUrl];
+};
+
+export default useFetch;
```
dennisoelkers: Can you please rename this file to `useFetch.js`? It's part of our implicit naming convention to name the file based on a default export, if one exists.

src/web/aws-cloudwatch/hooks/fetch.js
```diff
@@ -0,0 +1,81 @@
...
+USE:
+`const [status, setUrl] = useFetch(setNextHook, onFooBar);`
+ - `status` will provide the current reducer state `{ loading, error, data }`
+ - `setUrl` will be your hook to call as a submit handler within a subfunction, it needs the API route as a string:
+
+EXAMPLES:
+ - See uses in `src/web/aws-cloudwatch/StepAuthorize.jsx`
+*/
+
+const useFetch = (setHook, method = 'GET', options = {}, callback = () => {}) => {
```
dennisoelkers: Two things regarding the call signature, my proposal is this:

```
const useFetch = ({ method = 'GET', url = undefined, options = {} }, callback = () => {}) => {
  [...]
  const [fetchUrl, setFetchUrl] = useState(url);
```

  1.) I think we can collapse the `setHook` and the `callback`, they are doing the same, from `useFetch`'s perspective. If we want to have both a setter and a callback, we can call them sequentially from an anonymous function. (`callback` would get the state as parameter in this case)

  2.) If we are passing the `url` parameter when initializing `useFetch` (instead of through the returned setter), we can immediately do the fetch (instead of having to wrap the setter call in `useEffect` clause, like in `StepAuthorize`)

src/web/aws-cloudwatch/KinesisStreams.jsx
```diff
@@ -2,20 +2,39 @@ import React, { useContext } from 'react';
...
 import FormWrap from '../common/FormWrap';
 import ValidatedInput from '../common/ValidatedInput';
-import Routes from '../common/Routes';
+import Routes, { ApiRoutes } from '../common/Routes';
+import { renderOptions } from '../common/Options';
 
 const KinesisStreams = ({ onChange, onSubmit }) => {
   const { formData } = useContext(FormDataContext);
+  const { availableStreams, setLogSample } = useContext(ApiContext);
+  const [fetchStatus, setFetchOptions] = useFetch(
```
dennisoelkers: These need better names. What is actually fetched here? I guess it should be `[logSampleStatus, setLogSampleUrl]`?

PR #144, opened 2019-07-16
Use standard field names for AWS key and secret

danotorrey:
Use these standard key/secret field names for AWS key and secret for the Save Input AWS endpoint (From the AWSRequest interface). All other AWS API requests should be using these already.
`aws_access_key_id` (previously `aws_access_key`)
`aws_secret_access_key` (previously `aws_secret_key`)

The region field was also changed to use the central `AWSRequest` constant.


src/main/java/org/graylog/integrations/aws/resources/requests/AWSInputCreateRequest.java
```diff
@@ -29,15 +29,12 @@
 @JsonAutoDetect
 @AutoValue
 @WithBeanGetter
-public abstract class AWSInputCreateRequest {
+public abstract class AWSInputCreateRequest implements AWSRequest {
```
danotorrey: This implementation of the `AWSRequest` interface ensures that the key/secret/region fields are all specified. Most AWS API request objects implement this.

danotorrey:
@ceruleancee Feel free to merge this if it looks good to you.

src/main/java/org/graylog/integrations/aws/resources/requests/AWSInputCreateRequest.java
danotorrey: @kyleknighted We can help to cherry-pick this in to any branch that you are currently working on once merged.

PR #146, opened 2019-07-17
API Error Handling

### LINKED ISSUE #97: Error States
# Description

When the API sends back an Error, we need to handle it gracefully

kyleknighted:
Output the API errors in a collapsable container that shows a friendly message if available and the API response message as a fallback or as the collapsed additional info.

### All Steps
- [ ] ~Integrate `UserNotifications` from server~
- [x] Verify that errors only render once

### Auth Step
- [x] Invalid Key
- [x] Invalid Secret
- [x] Auth fallback error for uncaptured

### Streams
- [x] No Streams Found
- [ ] ~No Streams & No Group Found~ (to be handled in upcoming PR for Automated Setup)

### Health Check
- [x] No Logs found in Stream
- [x] Invalid Log format

### Review
- [x] Unable to Save Input

## Screenshots

ToggleApiError
image
image

Closes #97

### DIFF
--- a/src/main/java/org/graylog/integrations/aws/service/KinesisService.java
+++ b/src/main/java/org/graylog/integrations/aws/service/KinesisService.java
@@ -127,7 +127,7 @@ public class KinesisService {
         // Retrieve one records from the Kinesis stream
         final List<Record> records = retrieveRecords(request.streamName(), kinesisClient);
         if (records.size() == 0) {
-            throw new BadRequestException("The Kinesis stream does not contain any messages.");
+            throw new BadRequestException(String.format("The Kinesis stream [%s] does not contain any messages.", request.streamName()));
         }
 
         // Select random record from list, and check if the payload is compressed
--- a/src/main/java/org/graylog/integrations/aws/service/KinesisService.java
+++ b/src/main/java/org/graylog/integrations/aws/service/KinesisService.java
@@ -43,6 +43,7 @@ import software.amazon.awssdk.services.kinesis.model.Record;
 import software.amazon.awssdk.services.kinesis.model.ShardIteratorType;
 
 import javax.inject.Inject;
+import javax.ws.rs.BadRequestException;
 import javax.ws.rs.InternalServerErrorException;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -115,9 +116,7 @@ public class KinesisService {
         final boolean streamExists = kinesisStreamNames.streams().stream()
                                                        .anyMatch(streamName -> streamName.equals(request.streamName()));
         if (!streamExists) {
-            String explanation = String.format("The requested stream [%s] was not found.", request.streamName());
-            LOG.error(explanation);
-            return KinesisHealthCheckResponse.createFailed(explanation);
+            throw new BadRequestException(String.format("The requested stream [%s] was not found.", request.streamName()));
         }
 
         LOG.debug("The stream [{}] exists", request.streamName());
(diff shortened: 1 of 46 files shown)


src/main/java/org/graylog/integrations/aws/service/KinesisService.java
```diff
@@ -128,9 +127,7 @@ public KinesisHealthCheckResponse healthCheck(KinesisHealthCheckRequest request)
         // Retrieve one records from the Kinesis stream
         final List<Record> records = retrieveRecords(request.streamName(), kinesisClient);
         if (records.size() == 0) {
-            String explanation = "The Kinesis stream does not contain any messages.";
-            LOG.error(explanation);
-            return KinesisHealthCheckResponse.createFailed(explanation);
+            throw new BadRequestException("The Kinesis stream does not contain any messages.");
```
ceruleancee: Would be helpful to include the stream name. 
throw new BadRequestException(String.format("The Kinesis stream [%s] does not contain any messages.", request.streamName()));

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
```diff
@@ -243,10 +245,8 @@ private KinesisHealthCheckResponse handleCompressedMessages(KinesisHealthCheckRe
         Optional<CloudWatchLogEvent> logEntryOptional = data.logEvents().stream().findAny();
 
         if (!logEntryOptional.isPresent()) {
-            String message = "The CloudWatch payload did not contain any messages. This should not happen. " +
-                             "See https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html";
-            LOG.debug(message);
-            return KinesisHealthCheckResponse.createFailed(message);
+            throw new BadRequestException("The CloudWatch payload did not contain any messages. This should not happen. " +
```
ceruleancee: It would be cleaner to refactor this code block to a method in CloudWatchService that returns a CloudWatchLogEvent object and just check for the type of payload in healthCheck, should it change in the future
```
         // Assume that the payload is from CloudWatch.
        // Extract messages, so that they can be committed to journal one by one.
        final CloudWatchLogSubscriptionData data = objectMapper.readValue(bytes, CloudWatchLogSubscriptionData.class);

        if (LOG.isTraceEnabled()) {
            // Log the number of events retrieved from CloudWatch. DO NOT log the content of the messages.
            LOG.trace("[{}] messages obtained from CloudWatch", data.logEvents().size());
        }

        // Pick just one log entry.
        Optional<CloudWatchLogEvent> logEntryOptional = data.logEvents().stream().findAny();

        if (!logEntryOptional.isPresent()) {
            throw new BadRequestException("The CloudWatch payload did not contain any messages. This should not happen. " +
                                          "See https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html");
        }

        CloudWatchLogEvent logEntry = logEntryOptional.get();
```

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
danotorrey: Great idea! I'll fix.

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
danotorrey: Great point. I 100% agree. I actually have a method prepared in #151 that can be reused for this.

Once #151 is reviewed/merged, I will use it here too. I created issue https://github.com/Graylog2/graylog-plugin-integrations/issues/173 to track this 

https://github.com/Graylog2/graylog-plugin-integrations/blob/cf1cf124d141348d265b803f3897ffa2803f4c2e/src/main/java/org/graylog/integrations/aws/transports/KinesisPayloadDecoder.java#L95-L110

danotorrey:
This looks good to me.

@ceruleancee Thanks for the feedback. I have made the requested changes except for the one I opened https://github.com/Graylog2/graylog-plugin-integrations/issues/173 for.

PR #157, opened 2019-07-24
Remove emotion and use styled-components

kyleknighted:
This replaces Emotion with [styled-components](https://www.styled-components.com/) which i feel has better documentation and seems to have a more active community


danotorrey:
@kyleknighted Is there anything visible from this change that I can test in the UI?

kyleknighted:
@danotorrey It should look like nothing has changed

PR #160, opened 2019-07-24
Add addSubscriptionFilter to CloudwatchService

### LINKED ISSUE #62: Auto Kinesis setup part 5: Create the subscription
Programmatically create the CloudWatch log subscription using the policy created in #124. See these docs for reference.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Subscriptions.html
https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CreateSubscriptionFilter.html

ceruleancee:
Closes #62

### DIFF
--- a/src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
+++ b/src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
@@ -17,8 +17,6 @@ import javax.inject.Inject;
 import javax.ws.rs.BadRequestException;
 import java.util.ArrayList;
 
-import static java.lang.Thread.sleep;
-
 public class CloudWatchService {
 
     private static final Logger LOG = LoggerFactory.getLogger(CloudWatchService.class);
@@ -41,8 +39,8 @@ public class CloudWatchService {
      * Returns a list of log groups that exist in CloudWatch.
      *
      * @param region             The AWS region
-     * @param awsAccessKeyId
-     * @param awsSecretAccessKey The AWS region
+     * @param awsAccessKeyId     The AWS accessKey
+     * @param awsSecretAccessKey The AWS secretKey
      * @return A list of log groups in alphabetical order.
      */
     public LogGroupsResponse getLogGroupNames(String region, String awsAccessKeyId, String awsSecretAccessKey) {
@@ -68,8 +66,8 @@ public class CloudWatchService {
 
     public String addSubscriptionFilter(CloudWatchLogsClient cloudWatch, String logGroup, String streamArn,
                                         String roleArn, String filterName,
-                                        String filterPattern) throws InterruptedException {
-        PutSubscriptionFilterRequest putSubscriptionFilterRequest =
+                                        String filterPattern) {
+        final PutSubscriptionFilterRequest putSubscriptionFilterRequest =
                 PutSubscriptionFilterRequest.builder()
                                             .logGroupName(logGroup)
                                             .filterName(filterName)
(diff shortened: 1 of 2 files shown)


src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
```diff
@@ -54,4 +60,32 @@ public LogGroupsResponse getLogGroupNames(String region, String awsAccessKeyId,
...
+            final String responseMessage = String.format("Success. The subscription filter [%s] was added to [%s].",
+                                                         filterName, logGroup);
+            return responseMessage;
+        } catch (Exception e) {
+            final String specificError = ExceptionUtils.formatMessageCause(e);
+            final String responseMessage = String.format("Attempt to add subscription [%s] to Cloudwatch log group " +
+                                                         "[%s] failed due to the following exception: [%s]",
+                                                         filterName, logGroup, specificError);
+            LOG.error(responseMessage);
+            throw new InternalServerErrorException(responseMessage, e);
```
danotorrey: What do you think about throwing `javax.ws.rs.BadRequestException` here instead? `BadRequestException` will respond with a 400 HTTP code indicating that something is wrong with the request parameters. 

`InternalServerErrorException` returns a 500 HTTP code which generally says that it was the Graylog servers fault. 

I think the `BadRequestException` fits better, but happy to discuss more too. 

More info on statuses: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.1

src/main/java/org/graylog/integrations/aws/service/KinesisService.java
```diff
@@ -423,4 +423,5 @@ public KinesisNewStreamResponse createNewKinesisStream(KinesisNewStreamRequest k
             throw new InternalServerErrorException(responseMessage, e);
         }
     }
+
```
danotorrey: Can this extra line be removed?

src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
```diff
@@ -54,4 +60,32 @@ public LogGroupsResponse getLogGroupNames(String region, String awsAccessKeyId,
...
+        PutSubscriptionFilterRequest putSubscriptionFilterRequest =
+                PutSubscriptionFilterRequest.builder()
+                                            .logGroupName(logGroup)
+                                            .filterName(filterName)
+                                            .filterPattern(filterPattern)
+                                            .destinationArn(streamArn)
+                                            .roleArn(roleArn)
+                                            .distribution(Distribution.BY_LOG_STREAM)
+                                            .build();
+        sleep(2_000);
```
danotorrey: Is this sleep statement required?

src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
```diff
@@ -54,4 +60,32 @@ public LogGroupsResponse getLogGroupNames(String region, String awsAccessKeyId,
...
+        try {
+            cloudWatch.putSubscriptionFilter(putSubscriptionFilterRequest);
+            final String responseMessage = String.format("Success. The subscription filter [%s] was added to [%s].",
+                                                         filterName, logGroup);
+            return responseMessage;
+        } catch (Exception e) {
+            final String specificError = ExceptionUtils.formatMessageCause(e);
+            final String responseMessage = String.format("Attempt to add subscription [%s] to Cloudwatch log group " +
```
danotorrey: This `responseMessage` variable is not needed and can be in inline return statement like this:
```suggestion
            return String.format("Attempt to add subscription [%s] to Cloudwatch log group " +
```

src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
```diff
@@ -54,4 +60,32 @@ public LogGroupsResponse getLogGroupNames(String region, String awsAccessKeyId,
         LOG.debug("Log groups queried: [{}]", groupNameList);
         return LogGroupsResponse.create(groupNameList, groupNameList.size());
     }
+
+    public String addSubscriptionFilter(CloudWatchLogsClient cloudWatch, String logGroup, String streamArn,
+                                        String roleArn, String filterName,
+                                        String filterPattern) throws InterruptedException {
+        PutSubscriptionFilterRequest putSubscriptionFilterRequest =
```
danotorrey: The `putSubscriptionFilterRequest` variable can be finalized. 
```suggestion
        final PutSubscriptionFilterRequest putSubscriptionFilterRequest =
```

src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
ceruleancee: Good thinking. That exception makes more sense.

src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
ceruleancee: I kept the responseMessage and instead included `throw new BadRequestException(responseMessage, e);`

src/main/java/org/graylog/integrations/aws/service/CloudWatchService.java
ceruleancee: no its not. good catch! deleting.

Issue #29, opened 2019-04-30
Assume Role ARN Authentication


danotorrey:
Lennart mentioned that some customers using the existing AWS plugin used an alternate authentication scheme called Assume Role ARN. Apparently this was popular among bigger customers.

We will need to investigate how this works and see if we can support it in the new integration. Most likely this should just be a copy/paste operation from the existing AWS plugin.

From the existing code, it looks like it's just additional args provided during the auth provider setup. See https://github.com/Graylog2/graylog-plugin-aws/blob/ea7f5d91e7dbdfbe809d3cdc007bb5badbe69e79/src/main/java/org/graylog/aws/inputs/transports/KinesisTransport.java#L160 and the associated commit https://github.com/Graylog2/graylog-plugin-aws/commit/3c6ecf760f6ba36f2d87e4dedb0532ff898cd7ae

PR #188, opened 2019-08-13
Add support for assuming an ARN role

danotorrey:
## Overview 
Add ability to assume the ARN for another AWS account.

This follows the same patter as the older AWS plugin, but with the latest AWS SDK:

https://github.com/Graylog2/graylog-plugin-aws/blob/3c6ecf760f6ba36f2d87e4dedb0532ff898cd7ae/src/main/java/org/graylog/aws/auth/AWSAuthProvider.java

## Testing for this PR
1. Regression test to verify that the AWS Kinesis input has not broken with this change. Verify that the AWS Kinesis input starts successfully and pulls messages from Kinesis.

2. I am not sure how we can specifically this assume role functionality. (more info here). If the code looks good (if consistent with how assume role was done in the old AWS plugin) and this change does not break the input, I think it's ok to merge this PR. I have created #190 so we don't forget to revisit testing.


danotorrey:
Closes #29

src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java
```diff
@@ -140,6 +140,14 @@ public ConfigurationRequest combinedRequestedConfiguration() {
                     ConfigurationField.Optional.OPTIONAL,
                     TextField.Attribute.IS_PASSWORD));
 
+            request.addField(new TextField(
```
ceruleancee: delete unused `import org.graylog2.plugin.configuration.fields.BooleanField;`

src/main/java/org/graylog/integrations/aws/transports/KinesisConsumer.java
```diff
@@ -9,6 +9,7 @@
 import org.graylog2.plugin.system.NodeId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
```
ceruleancee: delete unused imports `org.graylog.integrations.aws.service.AWSService;` 
and `software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;`

PR #191, opened 2019-08-14
Kinesis auto-setup Health Check redirect

kyleknighted:
This PR properly redirects to the Health Check page and begins the API call to get the logs for recently setup stream subscriptions

### DIFF
--- a/src/web/aws-cloudwatch/KinesisSetup.jsx
+++ b/src/web/aws-cloudwatch/KinesisSetup.jsx
@@ -15,7 +15,7 @@ import { ApiContext } from './context/Api';
 
 import SetupModal from './auto-setup-steps/SetupModal';
 
-const KinesisSetup = ({ onChange, /* onSubmit, */ toggleSetup }) => {
+const KinesisSetup = ({ onChange, onSubmit, toggleSetup }) => {
   const { availableGroups, setGroups } = useContext(ApiContext);
   const { formData } = useContext(FormDataContext);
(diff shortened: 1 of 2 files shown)


src/web/common/Countdown.jsx
```diff
@@ -0,0 +1,63 @@
...
+  const [currentTime, setCurrentTime] = useState('00:00');
+  let logInterval;
+  let duration = timeInSeconds;
+
+  const startCountdown = () => {
+    let minutes;
+    let seconds;
+
+    logInterval = setInterval(() => {
+      minutes = parseInt(duration / 60, 10);
```
edmundoa: I'm wondering if it wouldn't be better to use a [Moment.js duration](https://momentjs.com/docs/#/durations/) to calculate and display the remaining time.

src/web/common/Countdown.jsx
```diff
@@ -0,0 +1,63 @@
...
+function Countdown({ callback, className, timeInSeconds, paused }) {
+  const [currentTime, setCurrentTime] = useState('00:00');
+  let logInterval;
+  let duration = timeInSeconds;
+
+  const startCountdown = () => {
+    let minutes;
+    let seconds;
+
+    logInterval = setInterval(() => {
```
edmundoa: Having a long timeout here, I wonder how this is going to work if the user switches tabs or windows in their computer. Can we still rely this timer will still be somewhat accurate? I know in the past there were throttling issues in some cases.

PR #194, opened 2019-08-14
Record when users agree to auto-setup

danotorrey:
Creates a record of when users agree to create AWS resources through the AWS Kinesis auto-setup flow. 

Since AWS resources cost money, we are explicitly recording when a particular user agrees to create resources though the Kinesis auto-setup flow.

Only the user id is logged (instead of the full user name) to avoid leaking user names/full names out into the server log file.

The resulting message looks like this (note that the local admin user name is logged since it does not actually exist in the database):
>  User [local:admin] agreed to the Kinesis auto-setup, which will create a Kinesis stream [test], role/policy, and a CloudWatch log group Subscription. This has been recorded, since AWS resources that will cost money are being created at the sole request of this user.

### DIFF
--- a/src/main/java/org/graylog/integrations/aws/resources/KinesisSetupResource.java
+++ b/src/main/java/org/graylog/integrations/aws/resources/KinesisSetupResource.java
@@ -17,7 +17,6 @@ import org.graylog.integrations.aws.service.CloudWatchService;
 import org.graylog.integrations.aws.service.KinesisService;
 import org.graylog2.plugin.database.users.User;
 import org.graylog2.plugin.rest.PluginRestResource;
-import org.graylog2.rest.resources.system.inputs.AbstractInputsResource;
 import org.graylog2.shared.rest.resources.RestResource;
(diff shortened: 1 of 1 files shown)


src/main/java/org/graylog/integrations/aws/resources/KinesisSetupResource.java
```diff
@@ -62,6 +64,15 @@ public KinesisSetupResource(CloudWatchService cloudWatchService, KinesisService
     public KinesisNewStreamResponse createNewKinesisStream(@ApiParam(name = "JSON body", required = true) @Valid @NotNull
                                                                    KinesisNewStreamRequest request) throws InterruptedException {
 
+        // Record the fact that a particular user agreed to create AWS resources.
```
ceruleancee: Maybe wording such as `"This has been recorded as the listed user has accepted the responsibility in potential incurring associated cost(s)."` 
instead of `"This has been recorded, since AWS resources that will cost money are being created at the sole request of this user"`

Issue #264, opened 2019-09-13
Support all AWS SDK authorization methods


danotorrey:
# Description

Fully support all AWS authorization methods for the new AWS Kinesis integration. Right now, an API key is required. The forces users to generate this extra credential set. It is sometimes more desirable to assume a specific role ARN or use the role of the EC2 instance.

## What

Support the following authorization schemes:
* Explicit credentials (key and secret - as we supported before).
* Automatic (supports environment variables, Java props, EC2 creds and other options described here).
* Assume Role ARN support

## Why

Assuming a role or policy in AWS is a common way to authorize applications running in the AWS environment. See https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html for more info. Requiring security keys add unneeded complexity and maintenance burdens for certain configurations.

danotorrey:
See https://support.graylog.com/#ticket/zoom/2597 for issue origin.

danotorrey:
The backend code for this is already in place, we just need to update the UI to provide the option to use API keys _or_ assume role ARN functionality.

danotorrey:
From talking to @kyleknighted about this, we think it's best to support all AWS SDK authorization options. It's not much more work to support them all, and it provides the most flexibility for our users.

So, on the first page of all AWS integrations, we would present the following three choices:

1) Explicit credentials (key and secret - as we supported before)

danotorrey:
From talking to @kyleknighted about this, we think it's best to support all AWS SDK authorization options. It's not much more work to support them all, and it provides the most flexibility for our users.

So, on the first page of all AWS integrations, we would present the following three choices:

1) Explicit credentials (key and secret - as we supported before).
2) Automatic (supports environment variables, Java props, EC2 creds and other options described [here](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/index.html?com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html)).
3) Assume Role ARN support.

This way, we support all possible options that could be needed since we are making changes to the auth code (not much more work to support them all).

PR #277, opened 2019-09-25
Moving CloudWatch and StepAuthorize

kyleknighted:
This PR prepares the Authorization step to be a shared component that multiple AWS services can use to authorize their account (for https://github.com/Graylog2/graylog-plugin-integrations/issues/264)

It is a precursor to allowing `ARN` and `Automatic` authorization options

Will be needed before we're able to build out the next AWS integration


danotorrey:
@kyleknighted I am looking at this now...

Issue #271, opened 2019-09-24
Support Interface VPC Endpoints for Kinesis Data Streams


kroepke:
# Description

A common pattern to avoid Kinesis traffic leaving a VPC is to use VPC endpoints instead of the public endpoint URLs.

## What

Allow to specify a private VPC endpoint URL instead of always using the public Kinesis endpoint URL.

## Why

If a user has set up their environments to use VPC endpoints, they cannot use the Graylog AWS plugin with it, because we don't allow overriding the Kinesis URL.
Follow https://github.com/awsdocs/amazon-kinesis-data-streams-developer-guide/blob/master/doc_source/vpc.md#using-interface-vpc-endpoints-for-kinesis-data-streams for more details.

danotorrey:
Clients that need support for Endpoints.
- CloudWatch
- DynamoDB
- IAM
- Kinesis

bernd:
This has been fixed in https://github.com/Graylog2/graylog-plugin-integrations/pull/279

PR #279, opened 2019-09-26
AWS Kinesis Enhancements

### LINKED ISSUE #264: Support all AWS SDK authorization methods

kyleknighted:
## Overview

This PR makes the following changes for the `graylog-plugin-integrations` AWS Kinesis/CloudWatch input:

1. Improve authorization support: Previously, only an explicit AWS Key/Secret could be used to authorize the input.  Now, the user can pick between using a key/secret and automatic AWS authorization, which uses the [DefaultAWSCredentialsProviderChain](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html), which allows many more flexible authentication options. 

image

See #264 for more info.

As part of this change, the Assume Role ARN option was also moved to the first page (previously was tucked-away on a subsequent page, which was too late, since many AWS SDK calls are initiate starting from the first page): 

image

2. Add ability to specify [VPC endpoints](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html) for all AWS API services used (CloudWatch, DynamoDB, IAM, and Kinesis). Fields for this are now present on the first page of the setup, since the endpoints must be used throughout the setup wizard. See #271 for more info.

image

Lots of code was changed to implement these, because many API calls support this functionality, and almost all of them had to be updated :)

Closes #264, #271


danotorrey:
I am planning to integrate the backend changes into this PR too. I'll start work on that tomorrow or Wednesday.

danotorrey:
@kyleknighted I have begun working on the backend for this. Sorry for the delay!

danotorrey:
@kyleknighted I noticed this error in the browser when running this branch:

> client.js:196 /Users/danieltorrey/workspace/graylog/master/graylog-project-repos/graylog-plugin-integrations/src/web/aws/common/awsAuth.jsModule not found: Error: Can't resolve './awsKeySecret' in '/Users/danieltorrey/workspace/graylog/master/graylog-project-repos/graylog-plugin-integrations/src/web/aws/common'

danotorrey:
@kyleknighted I have begun working on the backend for this again. I'l keep you updated on my progress.

danotorrey:
@kyleknighted Initial testing looks good so far. Noticed one possible issue:

Looks like the Assume Role ARN is not being passed in the Streams request (specified in input field, but is not included in the post request JSON payload):

image

danotorrey:
@kyleknighted The Endpoint support on the backend is ready for you. No rush in adding the frontend though. Next week is all good. 

Here's the wording for each of the endpoint text fields:
- CloudWatch API Endpoint Override
- DynamoDB API Endpoint Override
- IAM API Endpoint Override
- CloudWatch API  Endpoint Override

Here are the JSON request field names:
- `cloudwatch_endpoint`
- `dynamodb_endpoint`
- `iam_endpoint`
- `kinesis_endpoint`

Also, is there a way to work in a little info option (i) or or helper text that shows text like the following: "Overrides the AWS API endpoint URL that Graylog communicates with. Use this is you are using [VPC Endpoints](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html) for AWS services."

Once you get a chance to look at this, I am wondering if you can merge/rebase master again since lots has changed there recently.

danotorrey:
@kyleknighted The backend for Endpoint support that we talked about over Engineering Week is ready for you.

Here's the wording for each of the endpoint text fields:
- CloudWatch API Endpoint Override
- DynamoDB API Endpoint Override
- IAM API Endpoint Override
- CloudWatch API  Endpoint Override

Here are the JSON request field names:
- `cloudwatch_endpoint`
- `dynamodb_endpoint`
- `iam_endpoint`
- `kinesis_endpoint`

Also, is there a way to work in a little info option (i) or or helper text that shows text like the following: "Overrides the AWS API endpoint URL that Graylog communicates with. Use this is you are using [VPC Endpoints](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html) for AWS services."

danotorrey:
I will also update the AWS SDK to the latest with this PR too.

src/main/java/org/graylog/integrations/aws/AWSAuthProvider.java
```diff
@@ -0,0 +1,95 @@
...
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.auth.StsAssumeRoleCredentialsProvider;
+import software.amazon.awssdk.services.sts.model.AssumeRoleRequest;
+import software.amazon.awssdk.services.sts.model.GetCallerIdentityRequest;
+
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+
```
ceruleancee: delete extra line

src/main/java/org/graylog/integrations/aws/AWSAuthProvider.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Resolves the appropriate AWS authorization provider.
+ *
+ * If an {@code accessKey} and {@code secretKey} are provided, they will be used explicitly.
+ * If not, the default DefaultCredentialsProvider will be used instead. This will resolve the role/policy
+ * using using Java props, environment variables, EC2 instance roles etc. See the {@link DefaultCredentialsProvider}
```
ceruleancee: delete extra word `using`

src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
```diff
@@ -0,0 +1,101 @@
...
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClientBuilder;
+import software.amazon.awssdk.services.iam.IamClient;
+import software.amazon.awssdk.services.iam.IamClientBuilder;
+import software.amazon.awssdk.services.kinesis.KinesisClient;
+import software.amazon.awssdk.services.kinesis.KinesisClientBuilder;
+
+import java.net.URI;
+import java.net.URL;
```
ceruleancee: delete unneeded import

src/test/java/org/graylog/integrations/aws/service/KinesisServiceTest.java
```diff
@@ -312,6 +303,6 @@ public void testCreateNewKinesisStream() {
 
     @Test
     public void uniqueRoleName() {
-        assertEquals(1,1);
+        assertEquals(1, 1);
```
alex-konn: Seriously? ;)

danotorrey:
Some feedback from @alex-konn that I am planning to address in this PR:

### AWSAuthProvider
- [ ] Always created from request => add factory method to request? 
  `AWSRequest.toAuthProvider()` makes it easier to add additional auth
methods later and decouples other code from request fields that are only
auth-related

- [ ] Get rid of ACCESS_KEY prefix of role session name if none is provided
  or is it a CloudTrail convention to always have it?

### AWSClientBuilderUtil

- Could make this an instantiable class called e.g. `AWSClientFactory` 
  with aws builders constructor-injected (or constructed on-the-fly)
  that would remove the dependency on the builders from services using
clients

### KinesisConsumer

- [ ] `AWSRequest` ctor parameter is misleading. maybe rename the
  `AWSRequest` base class to `AWSParams` or something.
- [ ] could get its own config param object where the all the other ctor
  params are included too

danotorrey:
@kyleknighted Do you have some time to take a look at the yarn.lock merge conflict? I was not sure how to address that. :)

pom.xml
```diff
@@ -32,9 +32,8 @@
         <maven.deploy.skip>true</maven.deploy.skip>
         <maven.site.skip>true</maven.site.skip>
         <graylog.version>3.2.0-SNAPSHOT</graylog.version>
-        <aws-java-sdk-2.version>2.7.33</aws-java-sdk-2.version>
-        <aws-kinesis-client.version>2.2.7</aws-kinesis-client.version>
-        <integrations.protobuf.version>3.11.1</integrations.protobuf.version>
+        <aws-java-sdk-2.version>2.10.26</aws-java-sdk-2.version>
+        <aws-kinesis-client.version>2.2.6</aws-kinesis-client.version>
```
ceruleancee: Missing `<integrations.protobuf.version>3.11.1</integrations.protobuf.version>`

danotorrey: Thank you for catching this! I'll fix.

danotorrey: Fixed.

danotorrey:
As far as I am concerned, this PR is ready to be merged. I done some refactoring and cleanup. More can be done, but I think this is a good place to stop with this big PR:

- Remove unneeded and redundant comments.
- Don't include the ACCESS_KEY in the `roleSessionName` when not using key/secret.
- All redundant key/secret/endpoints/assumeRoleArn fields have been removed from request value objects. These objects still use AutoValue. The classes also now use the builder pattern. They are much more readable and maintainable now with no large `create()` methods with lots of arguments.

lingpri:
> @ceruleancee @lingpri @alex-konn
> As far as I am concerned, this PR is ready to be merged. I done some refactoring and cleanup. More can be done, but I think this is a good place to stop with this big PR:
> 
> * Remove unneeded and redundant comments.
> * Don't include the ACCESS_KEY in the `roleSessionName` when not using key/secret.
> * All redundant key/secret/endpoints/assumeRoleArn fields have been removed from request value objects. These objects still use AutoValue. The classes also now use the builder pattern. They are much more readable and maintainable now with no large `create()` methods with lots of arguments.
> 
> @kyleknighted Do you see any need for review of the UI changes?

lingpri:
> @ceruleancee @lingpri @alex-konn
> As far as I am concerned, this PR is ready to be merged. I done some refactoring and cleanup. More can be done, but I think this is a good place to stop with this big PR:
> 
> * Remove unneeded and redundant comments.
> * Don't include the ACCESS_KEY in the `roleSessionName` when not using key/secret.
> * All redundant key/secret/endpoints/assumeRoleArn fields have been removed from request value objects. These objects still use AutoValue. The classes also now use the builder pattern. They are much more readable and maintainable now with no large `create()` methods with lots of arguments.
> 
> @kyleknighted Do you see any need for review of the UI changes?

@danotorrey @kyleknighted @ceruleancee  Great work ! If you need any help with any specific task  Please keep me posted. I had a chance to look at creating the kinesis input  with Claudia's help and we are able to see the log messages. Also, we weren't sure how to test the VPC endpoints(Dynamo DB API ). Is there a Swagger documentation for these end points, that we can access. Thanks.

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
```diff
@@ -93,37 +92,63 @@ public void handleChangedThrottledState(boolean isThrottled) {
...
-        );
+        final String assumeRoleArn = configuration.getString(AWSInput.CK_ASSUME_ROLE_ARN);
+
+        final String dynamodbEndpoint = configuration.getString(AWSInput.CK_DYNAMODB_ENDPOINT);
+        final String cloudwatchEndpoint = configuration.getString(AWSInput.CK_CLOUDWATCH_ENDPOINT);
+        final String iamEndpoint = configuration.getString(AWSInput.CK_IAM_ENDPOINT);
+        final String kinesisEndpoint = configuration.getString(AWSInput.CK_KINESIS_ENDPOINT);
+
+        // Validate the endpoints, because the Kinesis client will silently fail if an endpoint is invalid.
+        validateEndpoint(dynamodbEndpoint, "DynamoDB");
```
lingpri: NB: Modify validateEndpoint(Map<K,V>) to keep calling the method multiple times.

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
```diff
@@ -93,37 +92,63 @@ public void handleChangedThrottledState(boolean isThrottled) {
...
-        }
-        this.kinesisConsumer = new KinesisConsumer(
-                nodeId, this, objectMapper, kinesisCallback(input), configuration.getString(CK_KINESIS_STREAM_NAME),
-                AWSMessageType.valueOf(configuration.getString(AWSCodec.CK_AWS_MESSAGE_TYPE)), region,
-                awsCredentialsProvider,
-                configuration.getInt(CK_KINESIS_RECORD_BATCH_SIZE, DEFAULT_BATCH_SIZE)
-        );
+        final String assumeRoleArn = configuration.getString(AWSInput.CK_ASSUME_ROLE_ARN);
+
+        final String dynamodbEndpoint = configuration.getString(AWSInput.CK_DYNAMODB_ENDPOINT);
```
lingpri: NB : Refactor
List<VPCEndpoints> lstOfVPCEndpoints =  configuration.awsService.getListOfVPCEndpoints() 
lstOfVPCEndpoints.forEach(x->validate(x))

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
```diff
@@ -93,37 +92,63 @@ public void handleChangedThrottledState(boolean isThrottled) {
...
 
+    /**
+     * If specified, verify that the specified endpoint is a valid URI.
+     *
+     * @param endpoint     The endpoint URI.
+     * @param endpointName The name of the endpoint. This will be included in the exception message.
+     * @throws MisfireException A {@link MisfireException} will be thrown if the URI is invalid.
+     */
+    private static void validateEndpoint(String endpoint, String endpointName) throws MisfireException {
+        if (StringUtils.isNotEmpty(endpoint)) {
```
lingpri: NB: Write a JUnit test to validate the endpoint , remove private block.

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
lingpri: @danotorrey After doing some slow thinking, if this feature is an optional one, we can create a new enhancement, to address this changes in the future release. ec2 describe-vpc-endpoints -> describes all the vpc endpoints that is available in a region. What are the default VPC endpoints, that we are configuring , if no options are provided ?

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
danotorrey: Thanks for the suggestion. I like the idea of using a loop instead of calling a method multiple time explicitly. 

The only way I can see that working here is to create a map, call `map.put`, then iterate or use `map.forEach` as you suggested. But, the downside is that it uses the same amount of code to accomplish the same result (due to map.put). I am not sure if that will actually improve the code here. I feel like we should leave it as is. Please feel free to provide a code sample if I missed anything here.

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
danotorrey: Great point. Will do.

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
danotorrey: I really like this enhancement. Perhaps we can consider this for the future.

src/main/java/org/graylog/integrations/aws/transports/KinesisTransport.java
danotorrey: Unit test added in https://github.com/Graylog2/graylog-plugin-integrations/pull/279/commits/8f341e6ac2abfda4d666124a1cce0a2798f18fce

danotorrey:
@lingpri @ceruleancee I investigated and found that there are valid use cases to allow an override endpoint for the IamClient AWS API URL. I feel that we should provide endpoint overrides for all AWS SDK services that we communicate to (just as the PR is currently coded to do).

* IAM endpoint override might be needed if the AWS FIPS endpoints are required by the user (https://aws.amazon.com/compliance/fips/)
* Other vendors that integrate with AWS allow IAM endpoint overrides. For example https://www.terraform.io/docs/providers/aws/guides/custom-service-endpoints.html#iam

Please also note that for the security concern of allowing IAM overrides, we are also adding URL [white list capability](https://github.com/Graylog2/graylog2-server/pull/6854), which allows the user to control which URLs are allowed to be used within Graylog. I will add support for this in a later PR.

danotorrey:
@ceruleancee @lingpri This PR is ready to be merged unless you have any other concerns.

src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
```diff
@@ -0,0 +1,100 @@
...
+import org.graylog.integrations.aws.resources.requests.AWSRequest;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.awscore.client.builder.AwsClientBuilder;
+import software.amazon.awssdk.core.client.builder.SdkClientBuilder;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClient;
+import software.amazon.awssdk.services.cloudwatchlogs.CloudWatchLogsClientBuilder;
+import software.amazon.awssdk.services.iam.IamClient;
+import software.amazon.awssdk.services.iam.IamClientBuilder;
+import software.amazon.awssdk.services.kinesis.KinesisClient;
```
lingpri: https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/examples-kinesis-stream.html
When do we use KinesisAsyncClient vs KinesisClient ?

src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java
```diff
@@ -145,6 +150,34 @@ public ConfigurationRequest combinedRequestedConfiguration() {
...
+                    ConfigurationField.Optional.OPTIONAL));
+
+            request.addField(new TextField(
+                    CK_DYNAMODB_ENDPOINT,
+                    "AWS DynamoDB Override Endpoint",
+                    "",
+                    AWS_SDK_ENDPOINT_DESCRIPTION,
+                    ConfigurationField.Optional.OPTIONAL));
+
+            request.addField(new TextField(
```
lingpri: @claudia - Do we want to keep this endpoint or remove IAM?

src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Initialize the builder with the appropriate authorization, region, and endpoints.
+     *
+     * @param builder  Any AWS client builder.
+     * @param endpoint See {@link SdkClientBuilder#endpointOverride(java.net.URI)} javadoc.
+     * @param region   The region to specify on the client.
+     */
+    public static void initializeBuilder(AwsClientBuilder builder, String endpoint, Region region, AwsCredentialsProvider credentialsProvider) {
+        builder.region(region);
```
lingpri: @ceruleancee Don't we have to perform null checks for endpoint or AWSCredentialProvider?

src/web/aws/authentication/AWSCustomEndpoints.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <StyledAdditionalFields title="Optional AWS VPC Endpoints"
+                            visible={isAWSCustomEndpointsVisible}
+                            onToggle={handleToggle}>
+
+      <StyledSectionTitle>Overrides the default AWS API endpoint URL that Graylog communicates with.</StyledSectionTitle>
+      <StyledSectionNote>Only specify these if you are using <ExternalLink href="https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints.html">VPC Endpoints</ExternalLink> for AWS services.</StyledSectionNote>
+
+      <ValidatedInput id="awsEndpointCloudWatch"
```
lingpri: @kyleknighted Are these inputs derived from the server side or they purely on the client side ?

src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java
danotorrey: We should keep it in my opinion. I have added more details in this comment:
https://github.com/Graylog2/graylog-plugin-integrations/pull/279#issuecomment-566003682

src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
danotorrey: @lingpri Great question. The synchronous Kinesis client is used in most of the API calls in [KinesisService](https://github.com/Graylog2/graylog-plugin-integrations/blob/91e8e1190b1e6c460a462977e9451b4451d71dbb/src/main/java/org/graylog/integrations/aws/service/KinesisService.java). We decided to use the sync client for all API requests here, because there these requests are blocking in the UI anyhow. The async client is used in [KinesisConsumer](https://github.com/Graylog2/graylog-plugin-integrations/blob/eb48aad6c9c1a90ba4c8832fe7a082162a0848ee/src/main/java/org/graylog/integrations/aws/transports/KinesisConsumer.java#L40) for reading messages from KinesisStreams in the normal runtime of the input. This usage requires async.

src/web/aws/authentication/AWSCustomEndpoints.jsx
danotorrey: @lingpri The value for these input fields is ultimately saved with the input that gets saved in Graylog. These input fields here are for the setup process, which uses the custom UI:

image

 The  input fields defined in [AWSInput](https://github.com/Graylog2/graylog-plugin-integrations/blob/628f89c10ade31d2de5bde1a8a53b19c731cb141/src/main/java/org/graylog/integrations/aws/inputs/AWSInput.java#L153) in Java code, are displayed in the Edit Input page modal dialog:

image

I hope this helps. Please let me know if you have any other questions.

src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
danotorrey: @lingpri Great point. I agree this is a good idea to do. I should probably do that consistently in the future. I like how the following example does this check (I believe this is a mostly standard way in Graylog code). What do you think?

It it is ok with you, I would like to wait on making this improvement, so it can be done later in a smaller single PR.

https://github.com/Graylog2/graylog2-server/blob/6fdb71e6b99c6b9fcad7d1853a64dfa6384a7609/graylog2-server/src/main/java/org/graylog2/cluster/ClusterConfigServiceImpl.java#L75

lingpri: @danotorrey -  I will have to slow think about this . Thanks for providing a good example . Yeah , please wait on making this improvement.

src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
lingpri: @danotorrey - Bingo . I was thinking about how to make the UI Non-blocking yesterday. I'm wondering if we can bring the **CLI/Terminal** in a wizard like interface.

PR #358, opened 2020-01-02
Fix for unable to launch IPFixUDPInput

lingpri:


src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,9 +82,9 @@
     protected IpfixCodec(@Assisted Configuration configuration, IpfixAggregator ipfixAggregator) throws IOException {
         super(configuration);
         this.ipfixAggregator = ipfixAggregator;
-        final URL standardIPFixDefTemplate = Resources.getResource("ipfix-iana-elements.json");
-        final String ipFixCustomDefPath    = (String) configuration.getSource().get(CK_IPFIX_DEFINITION_PATH);
-
+        //standard ipfix definition file is loaded from the classpath and is mandatory,we don't check for null or empty
```
danotorrey:
```suggestion
        // Standard IPFIX definition file is loaded from the classpath and is mandatory. We don't check for null or empty.
```
I am able to run with a production build and start IPFIX successfully too. Great work. 👍

This change looks good to me, but I am just suggesting some minor adjustments to the formatting of the comment.

danotorrey:
@lingpri It also looks like one of the commits in the PR came from one of your different user accounts. You might need to electronically sign the CLA from that user. 

image

Issue #350, opened 2019-12-20
IpfixUDP Input - Upload multiple custom definition templates


ceruleancee:
# Description
Enhance IpfixUDP input to support multiple custom definition files upload.

- implement Listfield as a configuration option 

- UI Enhancements - ability to add multiple file paths in the optional field

PR #359, opened 2020-01-02
Add listfield option for custom definition files

### LINKED ISSUE #353: Backend - Load ipfix custom definition files
## Description

As a developer I need to enhance ipfixcodec.java, and update unit test so that the functionality of loading multiple custom definition files is possible. 

Ref Parent : #350

ceruleancee:
Closes #353 and #350

### DIFF
--- a/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
+++ b/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
@@ -88,11 +88,9 @@ public class IpfixCodec extends AbstractCodec implements MultiMessageCodec {
     protected IpfixCodec(@Assisted Configuration configuration, IpfixAggregator ipfixAggregator) throws IOException {
         super(configuration);
         this.ipfixAggregator = ipfixAggregator;
-
         final URL standardIPFixDefTemplate = Resources.getResource(IpfixCodec.class, IPFIX_STANDARD_DEFINITION);
         final List<String> customDefFilePathList = configuration.getList(CK_IPFIX_DEFINITION_PATH);
         final List<URL> filePaths = new ArrayList<>();
-       
         if (customDefFilePathList == null || customDefFilePathList.isEmpty()) {
             infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate);
         } else {
--- a/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
+++ b/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
@@ -53,6 +53,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.URL;
@@ -60,9 +61,11 @@ import java.nio.file.Paths;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
@@ -85,14 +88,26 @@ public class IpfixCodec extends AbstractCodec implements MultiMessageCodec {
     protected IpfixCodec(@Assisted Configuration configuration, IpfixAggregator ipfixAggregator) throws IOException {
         super(configuration);
         this.ipfixAggregator = ipfixAggregator;
-        // Standard IPFIX definition file is loaded from the classpath and is mandatory. We don't check for null or empty.
         final URL standardIPFixDefTemplate = Resources.getResource(IpfixCodec.class, IPFIX_STANDARD_DEFINITION);
-        final String ipFixCustomDefPath = (String) configuration.getSource().get(CK_IPFIX_DEFINITION_PATH);
-        if (ipFixCustomDefPath == null || ipFixCustomDefPath.trim().isEmpty()) {
+        final List<String> customDefFilePathList = configuration.getList(CK_IPFIX_DEFINITION_PATH);
+        List<URL> validFilePathsList = new ArrayList<>();
+
+        if (customDefFilePathList == null || customDefFilePathList.isEmpty()) {
             infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate);
         } else {
-            final URL customIPFixDefURL = Paths.get(ipFixCustomDefPath).toUri().toURL();
-            infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate, customIPFixDefURL);
+            // TODO improve logic as needed and throw an error if file is not valid
+            for (String filePaths : customDefFilePathList) {
+                int index = 0;
+                File file = new File(customDefFilePathList.get(index));
+                if (validateFilePath(file)) {
+                    URL customDefURL = Paths.get(file.toURI()).toUri().toURL();
+                    validFilePathsList.add(customDefURL);
+                }
+            }
+            // TODO change hardcoded value
+            if (validFilePathsList.size() == 1) {
+                infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate, validFilePathsList.get(0));
+            }
         }
         this.parser = new IpfixParser(this.infoElementDefs);
     }
--- a/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
+++ b/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
@@ -117,11 +117,11 @@ public class IpfixCodec extends AbstractCodec implements MultiMessageCodec {
         if (customDefFile.isDirectory()) {
             String message = "The specified path is a folder. Please specify the full path to the file.";
             LOG.debug(message);
-            throw new IpfixException("An error occurred due to the following error [" + message+"]");
+            throw new IpfixException("An error occurred due to the following error [" + message + "]");
         } else if (!customDefFile.exists()) {
             String message = "The specified file [%s] does not exist.";
             LOG.debug(message);
-            throw new IpfixException("An error occurred due to the following error [" + message+"]");
+            throw new IpfixException("An error occurred due to the following error [" + message + "]");
         }
         return true;
     }
--- a/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
+++ b/src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
@@ -1,16 +1,16 @@
 /**
  * This file is part of Graylog.
- * <p>
+ *
  * Graylog is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * <p>
+ *
  * Graylog is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * <p>
+ *
  * You should have received a copy of the GNU General Public License
  * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
  */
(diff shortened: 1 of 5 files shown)


lingpri:
ipfix_custdefFile

@bernd @ceruleancee - I think, the validation for the field , needs to exist in the ListField and error messages rendered , as part of the error box under ListField component. Please advice.

bernd:
> @bernd @ceruleancee - I think, the validation for the custom definition field , needs to exist in the ListFieldComponent in the UI and error messages can be rendered , as part of the error box under ListField component. Please advice.
> 
> If we did server side validation, where will we show the error message, will it just fail to start the input ?

@lingpri Showing it under the field would be nice, but doesn't work with the current UI implementation of the inputs system. Server side validation errors will be shown in the red notification at the bottom of the page. That's the only thing we can use right now.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
         this.ipfixAggregator = ipfixAggregator;
-        // Standard IPFIX definition file is loaded from the classpath and is mandatory. We don't check for null or empty.
+
         final URL standardIPFixDefTemplate = Resources.getResource(IpfixCodec.class, IPFIX_STANDARD_DEFINITION);
-        final String ipFixCustomDefPath = (String) configuration.getSource().get(CK_IPFIX_DEFINITION_PATH);
-        if (ipFixCustomDefPath == null || ipFixCustomDefPath.trim().isEmpty()) {
+        final List<String> customDefFilePathList = configuration.getList(CK_IPFIX_DEFINITION_PATH);
+        List<URL> filePaths = new ArrayList<>();
+
+        if (customDefFilePathList == null || customDefFilePathList.isEmpty()) {
```
lingpri: @bernd @danotorrey - Below are some of my thoughts and wanted to share to bounce some ideas with you.
`Configuration.getURL(CK_IPFIX_DEFINITION_PATH)` can be a good  addition to the core code , that returns a 
`List<URL>`
L99-L108 can be refactored, with a code like below and would require us to slow think about this design.
```
ieds = Builder.build(InformationElementDefinitions.class)
                              .standardDef(standardIPFixDefTemplate)
                              .customDef(customDefFilePathList)
```

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
-            final URL customIPFixDefURL = Paths.get(ipFixCustomDefPath).toUri().toURL();
-            infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate, customIPFixDefURL);
+            errOnInvalidFilePath(customDefFilePathList);
+            filePaths.add(standardIPFixDefTemplate);
+            for (String filePath : customDefFilePathList) {
+                URL customDefURL = url(filePath.trim());
+                filePaths.add(customDefURL);
+            }
+            URL[] urls = array(filePaths);
+            infoElementDefs = new InformationElementDefinitions(urls);
```
bernd: What do you think about changing the constructor signature of `InformationElementDefinitions` to take a `List` of URLs instead of an array of URLs? That way we can avoid the whole list to array conversion and remove the `array()` method again.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
+        }
+    }
+
+    public void validateFilePath(File customDefFile) throws IpfixException {
+        if (customDefFile.isDirectory()) {
+            String message = "The specified path is a folder. Please specify the full path to the file.";
+            LOG.debug(message);
+            throw new IpfixException("An error occurred due to the following error [" + message + "]");
+        } else if (!customDefFile.exists()) {
+            String message = "The specified file [%s] does not exist.";
```
bernd: The format string (`%s`) in the message will not be substituted because we don't use `String.format()` here.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
+            File file = new File(filePath.trim());
+            validateFilePath(file);
+        }
+    }
+
+    public void validateFilePath(File customDefFile) throws IpfixException {
+        if (customDefFile.isDirectory()) {
+            String message = "The specified path is a folder. Please specify the full path to the file.";
+            LOG.debug(message);
+            throw new IpfixException("An error occurred due to the following error [" + message + "]");
```
bernd: I think using the `message` as exception message is enough. Adding "An error occurred ..." feels a bit redundant. :wink:

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
+        } else if (!customDefFile.exists()) {
+            String message = "The specified file [%s] does not exist.";
+            LOG.debug(message);
+            throw new IpfixException("An error occurred due to the following error [" + message + "]");
```
bernd: Same as above, using the `message` for the exception message should be enough.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -1,22 +1,22 @@
...
  * GNU General Public License for more details.
- * <p>
+ *
  * You should have received a copy of the GNU General Public License
  * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
  */
 package org.graylog.integrations.ipfix.codecs;
 
-
+import autovalue.shaded.com.google$.common.annotations.$VisibleForTesting;
```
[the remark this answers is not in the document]
ceruleancee: This has been removed

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
ceruleancee: Good point. 👍

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
ceruleancee: This has been removed

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -215,13 +260,15 @@ public void overrideDefaultValues(@Nonnull ConfigurationRequest cr) {
...
-                            "",
-                            "Path to the JSON file containing IPFIX field definitions",
-                            ConfigurationField.Optional.OPTIONAL)
-            );*/
+            configuration.addField(
+                    new ListField(CK_IPFIX_DEFINITION_PATH,
+                                  "IPFIX field definitions",
+                                  Collections.emptyList(),
+                                  Collections.emptyMap(),
+                                  "JSON file containing IPFIX field definitions",
```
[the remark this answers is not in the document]
ceruleancee: Good catch.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
+        for (String filePath : customDefFilePathList) {
+            File file = new File(filePath.trim());
+            validateFilePath(file);
+        }
+    }
+
+    public void validateFilePath(File customDefFile) throws IpfixException {
+        if (customDefFile.isDirectory()) {
+            String message = "The specified path is a folder. Please specify the full path to the file.";
+            LOG.debug(message);
```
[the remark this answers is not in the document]
ceruleancee: These have been removed

src/test/java/org/graylog/integrations/ipfix/codecs/IpfixAggregatorTest.java
```diff
@@ -78,6 +78,7 @@ public void multipleMessagesTemplateLater() throws IOException {
         assertThat(ipfixMessage.flows()).hasSize(4);
     }
 
+    @Ignore("Not ready.")
```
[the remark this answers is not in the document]
ceruleancee: Yes, the test fails now

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
lingpri: @bernd - can we take a safe approach, to have two constructors, one to take a List<URL's>? Also, I like arrays . I have an hunch, that arrays might be light weight than List. Please advice.

src/test/java/org/graylog/integrations/ipfix/codecs/IpfixAggregatorTest.java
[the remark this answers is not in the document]
lingpri: @danotorrey @ceruleancee - we need to look into this again. I'm surprised, its failing now.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +91,54 @@
...
-        if (ipFixCustomDefPath == null || ipFixCustomDefPath.trim().isEmpty()) {
+        final List<String> customDefFilePathList = configuration.getList(CK_IPFIX_DEFINITION_PATH);
+        List<URL> filePaths = new ArrayList<>();
+
+        if (customDefFilePathList == null || customDefFilePathList.isEmpty()) {
             infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate);
         } else {
-            final URL customIPFixDefURL = Paths.get(ipFixCustomDefPath).toUri().toURL();
-            infoElementDefs = new InformationElementDefinitions(standardIPFixDefTemplate, customIPFixDefURL);
+            errOnInvalidFilePath(customDefFilePathList);
```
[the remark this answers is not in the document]
lingpri: @danotorrey . I can change it to what you wish for. Also, It would be nice, if I can write something like this below.
      
```
.data(parse-config,ied,ip-fix-codec)
.err(bad-path,overload,interference,security-attack)
```

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
bernd: I would not use two constructors, that will complicate things. If you feel that arrays are better, then let's keep the arrays. :smiley: Arrays might be more lightweight, but in this case it's most probably not an issue.

lingpri: we are keeping arrays!!

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +88,52 @@
...
+    void checkValidFilePath(List<String> customDefFilePathList) throws IpfixException {
+        for (String filePath : customDefFilePathList) {
+            File file = new File(filePath.trim());
+            validateFilePath(file);
+        }
+    }
+
+    public void validateFilePath(File customDefFile) throws IpfixException {
+        if (customDefFile.isDirectory()) {
+            String message = "The specified path is a folder. Please specify the full path to the file.";
```
danotorrey: The `message` variable is not needed here. The string message can be explicitly provided within the `new IpfixException(message)`

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -82,18 +88,52 @@
...
+            throw new IpfixException(message);
+        } else if (!customDefFile.exists()) {
+            String message = "The specified file does not exist.";
```
danotorrey: Same as above. The `message` variable is not needed.

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -93,6 +93,9 @@ protected IpfixCodec(@Assisted Configuration configuration, IpfixAggregator ipfi
         final List<String> customDefFilePathList = configuration.getList(CK_IPFIX_DEFINITION_PATH);
         final List<URL> filePaths = new ArrayList<>();
 
+        //infoElementDefs = Builder.build(InformationElementDefinitions.class)
```
danotorrey: It looks like some extra commented out code was added in commit 584c4f1aadac4aa21aa74de6e9c1e77c0d4ba6b6. Can this be removed?

src/main/java/org/graylog/integrations/ipfix/codecs/IpfixCodec.java
```diff
@@ -108,6 +111,8 @@ protected IpfixCodec(@Assisted Configuration configuration, IpfixAggregator ipfi
         this.parser = new IpfixParser(this.infoElementDefs);
     }
 
+
```
danotorrey: Some extra blank lines were also added in commit 584c4f1aadac4aa21aa74de6e9c1e77c0d4ba6b6. Can this be removed?

PR #342, opened 2019-12-12
Retry when rate limit exception occurs for getStreams

danotorrey:
Fixes #341 

* Retries if Kinesis `LimitExceededException` occurs.
* Increases page size for streams from 30 to 400 to reduce the need for successive calls. Retrying successive requests will only be needed for extremely large numbers of streams.

**Ref :** 
    https://docs.aws.amazon.com/streams/latest/dev/fundamental-stream.html
    https://docs.amazonaws.cn/en_us/kinesis/latest/APIReference/kinesis-api.pdf
    https://docs.aws.amazon.com/cli/latest/reference/kinesis/create-stream.html

**CLI Reference Guide:**
https://docs.aws.amazon.com/cli/latest/reference/kinesis/describe-limits.html

**CLI-Example  for AWS services IAM(Dynamodb|Cloud Watch|Kinesis...):**
````
helium@helium-Precision-5540:~/workspace/config/shell$ aws iam get-user
{
    "User": {
        "UserName": "xxxx", 
        "PasswordLastUsed": "2019-12-12T17:00:43Z", 
        "CreateDate": "2019-11-06T14:01:44Z", 
        "UserId": "AIDAWV25I6RLUN4BMLKXB", 
        "Path": "/", 
        "Arn": "arn:aws:iam::459220251735:user/xxx"
    }
}
```

Issue #341, opened 2019-12-12
Kinesis Rate Limit exceeded


danotorrey:
## Description
When attempting to add a new Kinesis/CloudWatch input I am receiving an error after clicking "Authorize and Choose Streams". I can see this same error is logged the server.log, but no additional information before or after the error message itself. I am unable to find any AWS documentation on a limit related to Kinesis that we would potentially be hitting. Can you please advise? We are currently unable to add any new Kinesis inputs to Graylog.

> ERROR: software.amazon.awssdk.services.kinesis.model.LimitExceededException: Rate exceeded for account 943582610889. (Service: Kinesis, Status Code: 400, Request ID: dac85ac9-2965-5466-8ca3-756dbf1c2090)
## Steps To Reproduce

1. Create 200 or more streams in AWS.
1. Try to set up a new Kinesis integration.

## Environment

- Graylog Version: 3.1.3

danotorrey:
See ticket reference: https://support.graylog.com/#ticket/zoom/2998

PR #360, opened 2020-01-06
Backport - Retry when rate limit exception occurs for getStreams

danotorrey:
3.1 Backport for issue #342 and PR #341


danotorrey:
@lingpri @ceruleancee This is a backport for the Rate Limit issue #341 that was fixed previously. This backports the fix to the 3.1 branch, so that the 3.1.4 bug fix release will include the fix. The bug fix release is expected to be done in the second half of this week, so it would be awesome if you are able to merge this within the next days.

For this PR, I just directly cherry-picked commits 8b917cda5f58d9eb6d5f4921e1b804c57136396a 67914464e928c42aee0714fa25e1ecf3ae344e8d

Since commits were directly cherry-pickes, please just visually check to make sure that all fix code is there. For testing, just running a Graylog 3.1 branch checkout with this branch and verifying that you can progress through the AWS setup would be enough in my opinion. Of course, testing the issue scenario is great, but I don't think that is required, since it is so hard to cause, and the fix code is so concise here.

PR #462, opened 2020-04-28
Updates to fix AWS Input for GovCloud

waab76:
The AWS Input was not working for GovCloud regions due to hard-coded region value when constructing the IAM client.  Additionally, the recommended policy was missing a permission that is more strictly enfoced on GovCloud.  This updates IAM client construction to work for GovCloud and adds the missing permission to the recommended policy.

Tested locally with GovCloud Access Key and Secret.

### DIFF
--- a/src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
+++ b/src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
@@ -88,10 +88,14 @@ public class AWSClientBuilderUtil {
      * @return A fully built {@link IamClient}
      */
     public static IamClient buildClient(IamClientBuilder clientBuilder, AWSRequest request) {
+        Region iamRegion = Region.AWS_GLOBAL;
+        if (request.region().contains("gov")) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/graylog/integrations/aws/AWSClientBuilderUtil.java
```diff
@@ -88,10 +88,14 @@ public static KinesisClient buildClient(KinesisClientBuilder clientBuilder, AWSR
      * @return A fully built {@link IamClient}
      */
     public static IamClient buildClient(IamClientBuilder clientBuilder, AWSRequest request) {
+        Region iamRegion = Region.AWS_GLOBAL;
+        if (request.region().contains("gov")) {
```
danotorrey, line 92: This looks good to me. This seems like the best way to make the determination for which global region to use.

@waab76 Is it possible to add a test covering this if-statement?

waab76: Yeah, I'll do that now.

PR #464, opened 2020-04-29
GovCloud Fix

waab76:
Cherry-picked the commit for the GovCloud fix.  This change ensures that IAM client is constructed correctly for users setting up the Kinesis/Cloudwatch input to pull data from GovCloud.


danotorrey:
LGTM ✅

Issue #483, opened 2020-05-21
Support Palo Alto v9 Schema


waab76:
As a Graylog customer, I want the Palo Alto plugin to support the v9 schema so that I can make use of the new fields in that schema.

Requirements doc: https://graylogdocumentation.atlassian.net/wiki/spaces/GENERAL/pages/709165057/Palo+Alto+Input+Update

Input Criteria:
- [x] Palo Alto input works

Task List:
- [x] Build a sender script to enable testing
- [x] Mark existing Palo Alto input as "v8"
- [x] Create new Palo Alto input for v9
- [x] Create updated codecs to extract v9 schema fields for existing messages
- [x] Add new log types to the v9 codecs
- [x] New codecs use the Graylog Schema

Acceptance Criteria:
- [x] New Palo Alto input supports v9 schema
- [x] New Palo Alto input available in Graylog 4.0
- [x] New Palo Alto input available in Graylog 3.3

PR #485, opened 2020-05-22
Add a test script to send Syslog TCP/UDP messages to Graylog.

waab76:
In order to test Syslog-based plugins, we need the ability to generate traffic. For plugins like Palo Alto, where the traffic comes from a proprietary device, that can be quite difficult.  This script will send Syslog TCP/UDP messages with arbitrary payloads to a specified server/port, which allows us to take the payloads of real Palo logs and replay them as if they were coming from an actual Palo Alto device.

https://github.com/Graylog2/graylog-plugin-integrations/issues/483

Script usage:
Screen Shot 2020-05-22 at 8 54 20 AM

Standard Syslog TCP single message:
Screen Shot 2020-05-22 at 8 55 54 AM
Screen Shot 2020-05-22 at 8 56 31 AM

Standard Syslog TCP input file:
Screen Shot 2020-05-22 at 9 02 39 AM
Screen Shot 2020-05-22 at 9 03 45 AM

Palo Alto:
Screen Shot 2020-05-22 at 8 59 26 AM
Screen Shot 2020-05-22 at 8 59 57 AM

### DIFF
--- a/src/test/python/send_syslog.py
+++ b/src/test/python/send_syslog.py
@@ -1,0 +1,105 @@
+# This file is part of Graylog.
+#
+# Graylog is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Graylog is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


bud1979:
Works for me on my test system

PR #577, opened 2020-09-23
AWS CloudWatch Authentication Key & Secret Fix

kyleknighted:
`Key & Secret` were not being properly used during authentication, it was only trying to use the `Authomatic` authentication method of on-disk files.

This PR resolves the issue by placing a conditional within the fetch options.


waab76 (approved):
Pulled changes and tested locally.  Confirmed that the fix works.  Ship it!

PR #581, opened 2020-09-24
Reusing link component abstraction from core.

dennisoelkers:
This PR is reusing the link component abstraction from core, implemented in Graylog2/graylog2-server#9030.

/jenkins-pr-deps Graylog2/graylog2-server#9030


Gaya (approved):
Looks good. Just one file to actually update, but a lot of eslint fixed files.

PR #595, opened 2020-10-07
Slack plugin v001

lingpri:
1.   ported pr feed back [from](https://github.com/Graylog2/graylog2-server/pull/8893)
2.   Ability to send a slack message using graylog notification system.
3.   me and @waab76  have tested this locally in our machines.

### **Test Evidence**

slack_plugin


src/web/index.jsx
```diff
@@ -6,6 +6,8 @@ import Routes from 'aws/common/Routes';
 
 import AWSInputConfiguration from './aws/AWSInputConfiguration';
 import AWSCloudWatchApp from './aws/cloudwatch/CloudWatchApp';
+import SlackNotificationForm from './event-notifications/event-notification-types/SlackNotificationForm';
+import SlackNotificationSummary from './event-notifications/event-notification-types/SlackNotificationSummary';
```
waab76, line 10: As with the Java code, I don't think there's any reason to so closely follow the package structure from the graylog2-server repository.  I would suggest moving these to a folder simply called `slack`

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -0,0 +1,164 @@
...
+             + '${foreach event.fields field}  ${field.key}: ${field.value}${end}\n'
+             + '${if graylog_url}Graylog URL: ${graylog_url}\n${end}'
+             + '\n'
+             + '##########\n'
+             + '\n'
+             + '${if backlog}Last messages accounting for this alert:\n'
+             + '${foreach backlog message}${message}\n'
+             + '\n'
+             + '${end}${else}<No backlog>\n'
+             + '${end}',
```
waab76, line 36: I'm not so sure about this template.  I think it might be better to have something more like the email template in the documentation: https://docs.graylog.org/en/3.3/pages/alerts.html#email-alert-notification

waab76 (changes requested):

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: I will look into this.

src/web/index.jsx
lingpri: @waab76 - The suggestion came from @bernd to use a similar structure in the server repository.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,236 @@
...
+
+
+			final Notification systemNotification = notificationService.buildNow()
+					.addNode(nodeId.toString())
+					.addType(Notification.Type.GENERIC)
+					.addSeverity(Notification.Severity.NORMAL)
+					.addDetail("exception", exceptionDetail);
+			notificationService.publishIfFirst(systemNotification);
+
+			throw new PermanentEventNotificationException("Slack notification is triggered, but sending failed. " + e.getMessage(), e);
```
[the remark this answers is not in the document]
lingpri: @waab76 is this just a name change ?

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,236 @@
...
+		this.nodeId = requireNonNull(nodeId);
+		this.streamService = Optional.ofNullable(streamService);
+
+	}
+
+
+	@Override
+	public void execute(EventNotificationContext ctx) throws PermanentEventNotificationException {
+		final SlackEventNotificationConfig config = (SlackEventNotificationConfig) ctx.notificationConfig();
+		SlackClient slackClient = new SlackClient(config,okHttpClientProvider.get());
```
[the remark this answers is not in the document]
lingpri: yeah its a good one.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,236 @@
...
+								  NotificationService notificationService,
+								  OkHttpClientProvider okHttpClientProvider,
+								  NodeId nodeId, StreamService streamService){
+		this.notificationCallbackService = notificationCallbackService;
+		this.objectMapper = requireNonNull(objectMapper);
+		this.templateEngine = requireNonNull(templateEngine);
+		this.okHttpClientProvider = requireNonNull(okHttpClientProvider);
+		this.notificationService = requireNonNull(notificationService);
+		this.nodeId = requireNonNull(nodeId);
+		this.streamService = Optional.ofNullable(streamService);
```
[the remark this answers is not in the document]
lingpri: we do not want to deal with null checks. also, the streams is `optional` in the event definitions. I think, it should be `Optional`.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,236 @@
...
+	public interface Factory extends EventNotification.Factory {
+		@Override
+        SlackEventNotification create();
+	}
+
+
+	private static final Logger LOG = LoggerFactory.getLogger(SlackEventNotification.class);
+
+	private final EventNotificationService notificationCallbackService;
+	private final Optional<StreamService> streamService ;
```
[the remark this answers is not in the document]
lingpri: as I said below, the streams are optional from a functional standpoint and hence the dependency should be `Optional`

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	String buildCustomMessage(EventNotificationContext ctx, SlackEventNotificationConfig config, String template) {
+		List<MessageSummary> backlog = getAlarmBacklog(ctx);
+		Map<String, Object> model = getCustomMessageModel(ctx, config, backlog);
+		try {
+			return templateEngine.transform(template, model);
+		} catch (Exception e) {
+			LOG.error("Exception during templating", e);
+			return e.toString();
```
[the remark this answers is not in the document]
lingpri: yeah agree.

src/main/java/org/graylog/integrations/notifications/types/SlackClient.java
```diff
@@ -0,0 +1,159 @@
...
+
+	private final String webhookUrl;
+	private String proxyURL;
+	private final OkHttpClient httpClient;
+
+
+
+    @Inject
+	public SlackClient(SlackEventNotificationConfig configuration, OkHttpClient httpClient) {
+		this.webhookUrl = configuration.webhookUrl();
```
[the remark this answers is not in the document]
lingpri: @waab76 - please check with @bernd on this.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
waab76: I feel like you may be misunderstanding the purpose of `Optional` in Java.

`Optional` was added to the language to help avoid the pain of dealing with NullPointerExceptions.  If you are writing a method that returns an object but may sometimes return null, using an `Optional` as your return type will help the people who call your method (including you) gracefully handle the cases when it returns null instead of the expected object.  `Optional` is intended for use as a return type.  The only reason to declare something in your own code as `Optional` is if it is being populated by a method that returns an `Optional`.

In this case, you seem to be using it as a way to signal that this particular dependency may not be needed.  That's not how `Optional` is supposed to be used.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
waab76: See above comment.  `Optional` was added to Java to make it easier to handle instances where `null` may be returned instead of an object.  Guice is not going to inject a `null` value for StreamService.  Therefore there is no reason to use `Optional`.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
waab76: `PermanentEventNotificationException` and `TemporaryEventNotificationException` both extend `EventNotificationException`.  Like I said, I haven't checked to see if they are handled differently, but I would only throw a `PermanentEventNotificationException` if I knew there was a permanent failure.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -0,0 +1,167 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import lodash from 'lodash';
```
kyleknighted, line 3: lets not load the entire lodash library, only the functions you need
```suggestion
import cloneDeep from 'lodash/cloneDeep';
```

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -0,0 +1,167 @@
...
+import { Input } from 'components/bootstrap';
+import FormsUtils from 'util/FormsUtils';
+import { Button, ControlLabel, FormGroup, HelpBlock } from 'components/graylog';
+import { ColorPickerPopover } from 'components/common';
+import ColorLabel from 'components/sidecars/common/ColorLabel';
+
+class SlackNotificationForm extends React.Component {
+    static propTypes = {
+      config: PropTypes.object.isRequired,
+      validation: PropTypes.object.isRequired,
```
kyleknighted, line 14: Can we make these [`shape`](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes) props?

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -0,0 +1,167 @@
...
+      notify_channel: false,
+      link_names: false,
+      icon_url: '',
+      icon_emoji: '',
+
+    };
+
+    propagateChange = (key, value) => {
+      const { config, onChange } = this.props;
+      const nextConfig = lodash.cloneDeep(config);
```
kyleknighted, line 60:
```suggestion
      const nextConfig = cloneDeep(config);
```

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -0,0 +1,167 @@
...
+    render() {
+      const { config, validation } = this.props;
+
+      return (
+        <>
+
+          <FormGroup controlId="color">
+            <ControlLabel>Configuration color</ControlLabel>
+            <div>
+              <ColorLabel color={config.color || '#f06292'} />
```
kyleknighted, line 84: If `config` is required, then we should not need defaults here

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import CommonNotificationSummary from 'components/event-notifications/event-notification-types/CommonNotificationSummary';
+
+class SlackNotificationSummary extends React.Component {
+    static propTypes = {
+      type: PropTypes.string.isRequired,
+      notification: PropTypes.object,
+      definitionNotification: PropTypes.object.isRequired,
```
kyleknighted, line 10: Can we use `shape` here?

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    render() {
+      const { notification } = this.props;
+
+      return (
+        <CommonNotificationSummary {...this.props}>
+          <>
+            <tr>
+              <td>Color</td>
+              <td>{notification.config.color}</td>
```
kyleknighted, line 25: Since we do not have `notifications.config` required or defaulted, we should verify they are set
```suggestion
              <td>{notification?.config?.color}</td>
```

package.json
```diff
@@ -8,8 +8,12 @@
   },
   "scripts": {
     "build": "webpack",
-    "lint": "eslint --ext js,jsx src"
+    "lint": "eslint --ext js,jsx src/web --config ../graylog2-server/graylog2-web-interface/.eslintrc.json",
+    "lint:path": "eslint --ext js,jsx --config ../graylog2-server/graylog2-web-interface/.eslintrc.json",
+    "test": "jest",
+    "flow-check": "flow check"
   },
+
```
kyleknighted, line 16: Try out setting the `eslintConfig` and see if this still works as expected

```suggestion
    "lint": "eslint --ext js,jsx src/web",
    "lint:path": "eslint --ext js,jsx",
    "test": "jest",
    "flow-check": "flow check"
  },
  "eslintConfig": {
    "extends": [
      "../graylog2-server/graylog2-web-interface/.eslintrc.json"
    ]
  },
```

kyleknighted (changes requested):
Looks good, some comments can be applied to multiple similar concerns on the same component

package.json
```diff
@@ -18,7 +18,6 @@
     "graylog-web-plugin": "file:../graylog2-server/graylog2-web-interface/packages/graylog-web-plugin"
   },
   "private": true,
-  "devDependencies": {
-    "eslint-loader": "^1.6.3"
-  }
+  "devDependencies": {}
```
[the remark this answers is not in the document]
lingpri: @waab76 - I followed the intellij suggestion to resolve the error and later removed it , based on feedback from @kyleknighted . is there a problem in removing this dependency?

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: I will look into shape props.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: ok let me try that.

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
lingpri: i'm not familiar with shape .I will explore shape.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
lingpri: @waab76 I'm wondering how to simulate a template exception.

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.List;
+import java.util.Map;
+
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SlackEventNotificationTest extends SlackPluginTestFixture {
```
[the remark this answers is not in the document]
lingpri: the rest of the event notifications do not follow the factory pattern. we can explore this in the future. what do you think @waab76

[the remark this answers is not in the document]
waab76: None of the other event notifications need to get a new client object every time the notification config changes.  If they did, it would make sense to use a factory pattern for getting new client objects rather than calling the constructor directly.

I feel like this is worth addressing as part of this PR as it will let us have cleaner and clearer tests.  The `SlackEventNotification` code has no reason to know or care about the existence of OKHttp and yet, because of the way it is constructing `SlackClient` objects, if something related to OKHttp breaks, then the tests for `SlackEventNotification` will break as well.  The JUnit tests should only break when the class itself breaks, not when other classes break.

waab76 (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackClient.java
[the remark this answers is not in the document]
lingpri: i remember bernd suggesting, to follow the same repository as server.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
lingpri: @waab76 please check `buildCustomMessage` L147 .

package.json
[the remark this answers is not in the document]
kyleknighted: I don't recall where I mentioned changing dependencies, but I also don't see where this dep is even used in webpack in this repo.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
lingpri: i chose `temporaryexception` , since the error is a temporary glitch and can be quickly resolved by the user,

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import CommonNotificationSummary from 'components/event-notifications/event-notification-types/CommonNotificationSummary';
+
+class SlackNotificationSummary extends React.Component {
```
kyleknighted, line 6: This should be a function component instead of class - https://reactjs.org/docs/components-and-props.html#function-and-class-components

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import CommonNotificationSummary from 'components/event-notifications/event-notification-types/CommonNotificationSummary';
+
+function SlackNotificationSummary({ notification, ...restProps }) {
+  return (
+    <CommonNotificationSummary {...restProps} notification={notification}>
+      <>
```
kyleknighted, line 9: Unnecessary Fragment

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
```diff
@@ -0,0 +1,80 @@
...
+        </tr>
+        <tr>
+          <td>Icon Emoji</td>
+          <td>{notification.config.icon_emoji}</td>
+        </tr>
+        <tr>
+          <td>Graylog URL</td>
+          <td>{notification.config.graylog_url}</td>
+        </tr>
+      </>
```
kyleknighted, line 50: Unnecessary Fragment

kyleknighted (approved):
code lgtm with one minor note.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	String buildCustomMessage(EventNotificationContext ctx, SlackEventNotificationConfig config, String template) throws PermanentEventNotificationException {
+		List<MessageSummary> backlog = getAlarmBacklog(ctx);
+		Map<String, Object> model = getCustomMessageModel(ctx, config, backlog);
+		try {
+			LOG.debug("template = {} model = {}" ,template, model);
+			return templateEngine.transform(template, model);
+		} catch (Exception e) {
+			LOG.error("Exception during templating", e);
```
waab76, line 159: When you pass the Exception as an argument to `LOG.error()`, it will print the stack trace in the log.  As a general rule, we should only print the stack trace if we're not passing the exception back up the call chain because some method further up the call chain is also likely to print the stack trace and having multiple copies of the stack trace in the log is both inefficient and confusing for users.  Since this exception is going to be re-thrown, we should not print the stack trace.

```suggestion
			LOG.error("Exception during templating [{}]", e.toString());
```

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	@Override
+	@JsonIgnore
+	public ValidationResult validate() {
+		ValidationResult validation =  new ValidationResult();
+
+		final String regex = "https:\\/\\/hooks.slack.com\\/services\\/";
+		final Pattern pattern = Pattern.compile(regex, Pattern.MULTILINE);
+		final Matcher matcher = pattern.matcher(webhookUrl());
```
waab76, line 107: Having the regex `Pattern` be compiled every time `validate()` is called eliminates the efficiency of compiling the pattern.  The `regex` and `pattern` are not going to change from one call of `validate()` to another and thus you should consider promoting them up to class-level `static final` attributes

src/test/java/org/graylog/integrations/notifications/types/SlackClientTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    }
+
+    @After
+    public void tearDown() throws IOException {
+        server.shutdown();
+    }
+
+    @Test(expected = PermanentEventNotificationException.class)
+    public void send_message_with_okhttpclient_and_invalid_webhookurl() throws TemporaryEventNotificationException, PermanentEventNotificationException {
```
waab76, line 53: It's a good practice to name your test cases such that they include enough information to tell you what failed at a glance when looking at build output.  I like a pattern of `methodBeingTested_expectedBehavior_whenTheseConditionsArePresent()`.
```suggestion
    public void send_throwsException_whenWebhookUrlInvalid() throws Exception {
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfigTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class SlackEventNotificationConfigTest {
+
+    @Test
+    public void validate() {
```
waab76, line 16:
```suggestion
    public void validate_succeeds_whenWebhookUrlIsValid() {
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfigTest.java
```diff
@@ -0,0 +1,35 @@
...
+        SlackEventNotificationConfig slackEventNotificationConfig = SlackEventNotificationConfig.builder()
+                .webhookUrl("https://hooks.slack.com/services/xxxx/xxxx/xxxxxxx")
+                .build();
+        ValidationResult result = slackEventNotificationConfig.validate();
+        Map errors = result.getErrors();
+        assertThat(errors).size().isEqualTo(0);
+    }
+
+    @Test
+    public void test_invalid_slackwebhook() {
```
waab76, line 26:
```suggestion
    public void validate_failsAndReturnsAnError_whenWebhookUrlIsInvalid() {
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
+package org.graylog.integrations.notifications.types;
+
+import com.floreysoft.jmte.Engine;
+import com.google.common.collect.ImmutableList;
+import org.graylog.events.notifications.*;
```
waab76, line 5: Prefer specific imports to `*`

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfigTest.java
```diff
@@ -0,0 +1,35 @@
+package org.graylog.integrations.notifications.types;
+
+import org.graylog2.plugin.rest.ValidationResult;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.*;
```
waab76, line 9: Prefer specific imports

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.List;
+import java.util.Map;
+
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SlackEventNotificationTest extends SlackPluginTestFixture {
```
waab76, line 33: Now that the dependencies have been dealt with, `SlackEventNotification` should no longer be aware of OkHttp and this test should no longer need to extend `SlackPluginTestFixture`

```suggestion
public class SlackEventNotificationTest {
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Mock
+    NodeId mockNodeId;
+    private SlackEventNotification slackEventNotification;
+    private SlackEventNotificationConfig slackEventNotificationConfig;
+    private EventNotificationContext eventNotificationContext;
+
+    public SlackEventNotificationTest() throws IOException {
+    }
```
waab76, line 42: Don't think a no-arg constructor is needed.

```suggestion
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SlackEventNotificationTest extends SlackPluginTestFixture {
+
+    @Mock
+    NodeId mockNodeId;
```
waab76, line 36: All of the constructor parameters for `SlackEventNotification` should be here as Mocks
```suggestion
    @Mock 
    EventNotificationService mockEventNotificationService;
    @Mock
	ObjectMapper mockObjectMapper;
	@Mock
	Engine mockTemplateEngine;
	@Mock
	NotificationService mockNotificationService;
    @Mock
    NodeId mockNodeId;
	@Mock
	SlackClient mockSlackCLient;
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+    private SlackEventNotification slackEventNotification;
```
waab76, line 37: Use `@InjectMocks` so you don't need to manually build the test object in a `@Before` method

```suggestion
    @InjectMocks
    private SlackEventNotification slackEventNotification;
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+    private SlackEventNotificationConfig slackEventNotificationConfig;
+    private EventNotificationContext eventNotificationContext;
```
waab76, line 39: Both of these can also be declared with the `@Mock` annotation to avoid the need to manually create them as mocks.

src/test/java/org/graylog/integrations/notifications/types/SlackPluginTestFixture.java
```diff
@@ -0,0 +1,49 @@
...
+                null);
+
+        return provider;
+    }
+
+    List<String> getJsonNodeFieldValue(String expected,String fieldName) throws IOException {
+        final byte[] bytes = expected.getBytes();
+        JsonNode jsonNode = new ObjectMapper().readTree(bytes);
+        return jsonNode.findValuesAsText(fieldName);
+    }
```
waab76, line 47: This is only used in `SlackMessageTest`.  Why not move the method to that class?

src/test/java/org/graylog/integrations/notifications/types/SlackPluginTestFixture.java
```diff
@@ -0,0 +1,49 @@
...
+    OkHttpClientProvider getOkHttpClientProvider() {
+        final OkHttpClientProvider provider = new OkHttpClientProvider(
+                Duration.milliseconds(100L),
+                Duration.milliseconds(100L),
+                Duration.milliseconds(100L),
+                server.url("/").uri(),
+                null);
+
+        return provider;
+    }
```
waab76, line 41: Why are these two separate methods if `getOkHttpClientProvider()` is only ever called by `getOkHttpClient()`?

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+        final ImmutableList<MessageSummary> messageSummaries = ImmutableList.of(
+                new MessageSummary("graylog_1", new Message("Test message 1", "source1", new DateTime(2020, 9, 6, 17, 0, DateTimeZone.UTC))),
+                new MessageSummary("graylog_2", new Message("Test message 2", "source2", new DateTime(2020, 9, 6, 17, 0, DateTimeZone.UTC)))
+        );
+        EventNotificationService notificationCallbackService = mock(EventNotificationService.class);
+        when(notificationCallbackService.getBacklogForEvent(eventNotificationContext)).thenReturn(messageSummaries);
+
+        NotificationService mockNotificationService = mock(NotificationService.class);
+        SlackClient slackClient = new SlackClient(getOkHttpClient());
```
waab76, line 59: As mentioned above, you should be mocking the `SlackClient` for this test.

waab76 (changes requested):
Biggest thing I think needs to be fixed before merging to `master` is the redundant OkHttp binding in the module.  But while that's getting fixed, might as well look at some of the other (generally minor/style) things as well.

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
lingpri: i will give it a shot. probably, if it ends up having to remove the base class.

src/test/java/org/graylog/integrations/notifications/types/SlackPluginTestFixture.java
lingpri: @waab76 - Do you think I should mock `OkHttpClient` ?

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
lingpri: @waab76 i'm repurposing the pre-existing objects that were created earlier in the email and http notifications.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,179 @@
...
+
+		try {
+			SlackMessage slackMessage = createSlackMessage(ctx, config);
+			slackClient.send(slackMessage,config.webhookUrl());
+		} catch (Exception e) {
+			final Notification systemNotification = notificationService.buildNow()
+					.addNode(nodeId.toString())
+					.addType(Notification.Type.GENERIC)
+					.addSeverity(Notification.Severity.NORMAL)
+					.addDetail("exception", e.toString());
```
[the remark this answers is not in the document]
lingpri: I'm looking to add `LOG.error` and watch the server logs.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,179 @@
...
+
+	String buildCustomMessage(EventNotificationContext ctx, SlackEventNotificationConfig config, String template) throws PermanentEventNotificationException {
+		List<MessageSummary> backlog = getAlarmBacklog(ctx);
+		Map<String, Object> model = getCustomMessageModel(ctx, config, backlog);
+		try {
+			LOG.debug("template = {} model = {}" ,template, model);
+			return templateEngine.transform(template, model);
+		} catch (Exception e) {
+			LOG.error("Exception during templating", e);
+			throw new PermanentEventNotificationException(e.toString(),e.getCause());
```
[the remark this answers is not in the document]
lingpri: this needs some thought.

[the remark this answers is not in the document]
lingpri: the test `buildCustomMessage_with_invalidTemplate` does throw `PermanentEventNotificationException` . @waab76 is the test incorrect ?

src/test/java/org/graylog/integrations/notifications/types/SlackPluginTestFixture.java
lingpri: is there an issue in the logic?

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
lingpri: I have adjusted my intellij settings to apply the preferences.:)

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -0,0 +1,185 @@
...
+	 */
+	@Override
+	public void execute(EventNotificationContext ctx) throws PermanentEventNotificationException,TemporaryEventNotificationException {
+		final SlackEventNotificationConfig config = (SlackEventNotificationConfig) ctx.notificationConfig();
+		ValidationResult  result = config.validate();
+		result.getErrors().entrySet().stream().forEach(e -> LOG.error("Invalid configuration for key [{}] and value [{}]",e.getKey() , e.getValue()));
+
+		if(result.failed()){
+			throw new PermanentEventNotificationException("Please verify your Slack Event Configuration");
+		}
```
waab76, line 93: Config validation happens when a user creates/modifies a Notification from the UI. It should not be possible for invalid config to exist at this point in processing.
```suggestion
```

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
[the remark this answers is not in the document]
waab76: Your test only covers the behavior of the `buildCustomMessage` method.  What I'm saying is that `buildCustomMessage` throws a `PermanentEventNotificationException` which then gets caught by the try/catch block in `execute` and replaced with a `TemporaryEventNotificationException`.  If you believe it is right to throw a `PermanentEventNotificationException` here, then your try/catch block in `execute` should specifically catch `PermanentEventNotificationException` and re-throw it rather than wrapping it in a `TemporaryEventNotificationException`.

waab76 (approved):
No blocking issues remain, though I do think the JUnit tests need to be cleaned up/better organized.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
lingpri: https://github.com/Graylog2/graylog-plugin-integrations/issues/611 - i have created a new issue for this commit suggestion.

PR #606, opened 2020-10-21
Bring PagerDutyNotification into official integrations repo

waab76:
This PR migrates the Pager Duty plugin from Graylog Labs into core Graylog.

In addition to bringing PagerDuty into the integrations repo, this change also includes some refactoring to separate the PagerDutyClient from EventNotification concerns and to update the PagerDutyClient to use OkHttp rather than Apache HTTP.

This was tested locally on my machine by setting up PagerDuty notifications on an existing alert.

PagerDuty Notification setup UI:
Screen Shot 2020-10-27 at 5 24 18 PM

Notification in PagerDuty:
Screen Shot 2020-10-27 at 5 25 37 PM

Event details in PagerDuty:
Screen Shot 2020-10-27 at 5 26 22 PM

Test with invalid router key:
Screen Shot 2020-10-27 at 5 27 18 PM


src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
```diff
@@ -0,0 +1,107 @@
...
+    @Inject
+    public PagerDutyClient(final OkHttpClient httpClient,
+                    final ObjectMapper objectMapper,
+                    final MessageFactory messageFactory) {
+        this.httpClient = httpClient;
+        this.objectMapper = objectMapper;
+        this.messageFactory = messageFactory;
+    }
+
+    public PagerDutyResponse trigger(EventNotificationContext ctx) throws PagerDutyClientException {
```
lingpri, line 70: @waab76 - shouldn't this use `TemporaryEventNotificationException` and `PermanentEventNotificationException`

lingpri (changes requested):
i'm wondering, if the trigger method should throw  `TemporaryEventNotificationException` and 
`PermanentEventNotificationException` .

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
```diff
@@ -0,0 +1,107 @@
...
+        } catch (IOException e) {
+            throw new IllegalStateException("There was an error sending the notification event.", e);
+        }
+    }
+
+    private String buildRequestBody(EventNotificationContext ctx) throws PagerDutyClientException {
+        try {
+            return objectMapper.writeValueAsString(messageFactory.createTriggerMessage(ctx));
+        } catch (IOException e) {
+            throw new PagerDutyClientException("Failed to build trigger message", e);
```
lingpri, line 94: @waab76 - this should throw `TemporaryEventNotificaionException`

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    public static class PagerDutyClientException extends Exception {
```
lingpri, line 98: we can remove this `PagerDutyClientException`

lingpri (commented):
lgtm

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
waab76: I didn't want to use notification-specific exceptions in the PagerDutyClient because it could be used for other things in the future.  There is no reason to directly link the PagerDutyClient to our Event Notification system by using Event Notification exceptions here.

src/main/java/org/graylog/integrations/pagerduty/PagerDutyNotification.java
```diff
@@ -0,0 +1,67 @@
...
+            List<String> errors = response.getErrors();
+            if (errors != null && errors.size() > 0) {
+                throw new IllegalStateException(
+                        "There was an error triggering the PagerDuty event, details: " + errors);
+            }
+        } catch (PagerDutyClientException e) {
+            throw new EventNotificationException (
+                    "There was an exception triggering the PagerDuty event.", e);
+        } catch (Exception e) {
+            throw new EventNotificationException (
```
lingpri, line 62: `final Notification systemNotification = notificationService.buildNow()
					.addNode(nodeId.toString())
					.addType(Notification.Type.GENERIC)
					.addSeverity(Notification.Severity.NORMAL)
					.addDetail("SlackEventNotification send error ", e.toString());

			notificationService.publishIfFirst(systemNotification);`

The above notification should be published here isn't ? we follow the pattern in `SlackPlugin` . how ever i donot know what is the purpose of this . Any thoughts on this.

lingpri (changes requested):
`notificationService.publishIfFirst(systemNotification);` - again a design question, should the `systemnotification` be raise in the exception block . please advice.

src/main/java/org/graylog/integrations/pagerduty/PagerDutyNotification.java
waab76: I'll be honest, I haven't done a deep dive on error handling in this plugin yet.  My thinking is that we should raise a notification on the Graylog UI only in situations where user action is required to make the plugin work (the same situations where we would want to throw a `PermanentEventNotificationException`) but not in situations where the problem might be transient.

So if the user has entered an invalid PagerDuty key, then we might want to throw a `PermanentEventNotificationException` and raise a notification on the UI.  But if there was a network timeout trying to do the POST to the PagerDuty API, then we might want to throw a `TemporaryEventNotificationException` and not raise a notification on the UI.

Looking at the HTTP Notification and Email Notification, the HTTP Notification never uses UI notifications while the Email Notification does.

waab76: Fun fact I just learned: PagerDuty doesn't seem to do any validation of the `Routing Key`.  If the user puts in a bad routing key, they just won't get the notification in PagerDuty.

lingpri: so yeah repeating the question again, do we need `NotificationService` as a dependency in this class and do we call ,
`notificationService.publishIfFirst(systemNotification);` in this exception block ? probably @bernd can help with our queries.

lingpri: > Fun fact I just learned: PagerDuty doesn't seem to do any validation of the `Routing Key`. If the user puts in a bad routing key, they just won't get the notification in PagerDuty.

validation is not `critical` and can be addressed later. :)

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
```diff
@@ -0,0 +1,116 @@
...
+                } else if (429 == response.code()) {
+                    throw new TemporaryPagerDutyClientException("Too many PagerDuty API calls at one time");
+                } else {
+                    throw new TemporaryPagerDutyClientException(String.format(
+                            "HTTP %d - PagerDuty server encountered an internal error", response.code()));
+                }
+            }
+            return objectMapper.readValue(response.body().string(), PagerDutyResponse.class);
+        } catch (IOException e) {
+            throw new TemporaryPagerDutyClientException("There was an error sending the notification event.", e);
```
lingpri, line 97: add `log.error` here

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
```diff
@@ -0,0 +1,116 @@
...
+            throws TemporaryPagerDutyClientException, PermanentPagerDutyClientException {
+        final Request request = new Request.Builder()
+                .url(API_URL)
+                .post(RequestBody.create(MediaType.parse(APPLICATION_JSON), payloadString))
+                .build();
+
+        LOG.debug("Triggering event in PagerDuty with POST payload: {}", payloadString);
+
+        try (final Response response = httpClient.newCall(request).execute()) {
+            if (!response.isSuccessful()) {
```
lingpri, line 83: please add `log.debug` to print essential elements like `response headers` ,`status code`, `response size`, total time taken to get the response back etc.

src/main/java/org/graylog/integrations/pagerduty/client/MessageFactory.java
```diff
@@ -0,0 +1,128 @@
...
+                config.routingKey(),
+                "trigger",
+                dedupKey,
+                config.clientName(),
+                config.clientUrl(),
+                streamLinks,
+                payload);
+    }
+
+    private Link buildStreamWithUrl(Stream stream, EventNotificationContext ctx, PagerDutyNotificationConfig config) {
```
lingpri, line 108: do we need this method ?

src/main/java/org/graylog/integrations/pagerduty/PagerDutyNotification.java
```diff
@@ -0,0 +1,106 @@
...
+        } catch (PagerDutyClient.TemporaryPagerDutyClientException e) {
+            throw new TemporaryEventNotificationException(
+                    String.format("Error enqueueing the PagerDuty event :: %s", e.getMessage()),
+                    null != e.getCause() ? e.getCause() : e);
+        } catch (PagerDutyClient.PermanentPagerDutyClientException e) {
+            String errorMessage = String.format("Error enqueueing the PagerDuty event :: %s", e.getMessage());
+            Notification systemNotification = notificationService.buildNow()
+                    .addNode(nodeId.toString())
+                    .addType(Notification.Type.GENERIC)
+                    .addSeverity(Notification.Severity.URGENT)
```
lingpri, line 87: how do we determine the severity is `urgent` or `normal`?

lingpri (commented):
1. please verify again if the `mandatory` and `optional` fields belong in the right bucket.most of the fields seem to be mandatory.
2.  add a few additional log.debug and log.error statements.
3.  the client code doesn't implement an interface, so there's a diversity of method signatures, some return `response`, some `void` and with their own individualistic exception signatures. probably, we need to discuss, if this needs to be standardized if a pattern is emerging across multiple `client` code.
4. Great work on the PR.

src/main/java/org/graylog/integrations/pagerduty/PagerDutyNotification.java
waab76: Generally speaking, if you are sending something to PagerDuty it is because you want your ops people to be aware of it immediately.  Therefore, if you have a problem that will prevent anything from being sent to PagerDuty, that is an urgent issue that requires immediate attention.

src/main/java/org/graylog/integrations/pagerduty/client/MessageFactory.java
waab76: It is being used.  I kept it because the intent here is to bring the PagerDuty Notification from the Graylog Labs repo into our official release without changing the functionality.

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
waab76: Will do.

src/main/java/org/graylog/integrations/pagerduty/client/PagerDutyClient.java
waab76: Will do.

src/web/pager-duty/PagerDutyNotificationSummary.jsx
```diff
@@ -0,0 +1,49 @@
...
+      custom_incident: PropTypes.bool,
+      key_prefix: PropTypes.string,
+      client_name: PropTypes.string,
+      client_url: PropTypes.string,
+    }).isRequired,
+  }).isRequired,
+  definitionNotification: PropTypes.shape.isRequired,
+};
+
+PagerDutyNotificationSummary.defaultProps = {};
```
kyleknighted, line 47: Unnecessary but not a deal break

waab76: Removed

kyleknighted (approved):
frontend code lgtm

cobaltclaudia:
per your demo, and simple local test working as expected, lgtm 👍 
Screen Shot 2020-10-29 at 12 11 17 PM

cobaltclaudia (approved):
LGTM

PR #617, opened 2020-10-27
Remove model data

lingpri:
- remove model data.


waab76:
I'm not sure what this PR is for.  The title says "remove model data" but this appears to be adding 3 files of model data.  Please clarify the purpose of this PR.

Issue #623, opened 2020-10-29
Slack Plugin should support Backlog Message Limit


lingpri:
Slack Notification users may want to limit the number of backlog messages sent in a Slack notification in order to prevent flooding their channels with data.  To support this, we will add a config option to Slack Notification to set a backlog message limit.  When set, this should ensure that no more than the specified number of backlog messages are sent to Slack.

As an example, a user could set their Alert to include 50 Backlog Messages but they may only want 3 messages to be included in the Slack Notification.

# Input Criteria
- [x] Slack plugin works

# Output Criteria
- [x] Slack Notification has configuration options to allow users to set a Backlog Message Limit
- [x] Text on the UI makes it clear that the Slack Notification setting will override Alert settings
- [x] When no Backlog Message Limit is set, Slack Notifications send all included Backlog Messages to Slack
- [x] When the Backlog Message Limit is set, Slack Notifications do not send more than the specified number of Backlog Messages to Slack
- [x] Slack Notifications default to limit 0 in backend and frontend
- [x] Users are able to disable the limit after it has been enabled

# Tasks
- [x] Determine the best way to represent the limit in code
- [x] Update config classes to store all data related to the limit
- [x] Update UI to enable users to enter limit
- [x] Update Slack logic to enforce the limit
- [x] JUnit tests written to validate that the limit is enforced
- [x] Test locally
- [x] Slack Notification documentation updated to describe the additional Message Backlog Limit

# Improvements
- [x] Fix help block for custom message and error message description (https://github.com/Graylog2/graylog-plugin-integrations/pull/639/commits/5405878fa0409d8ee50ade8ac3a4e74453c7ade1)

# Reference
- [ ] https://github.com/graylog-labs/graylog-plugin-slack/tree/master/src/main

lingpri:
![backlog_items](https://user-images.githubusercontent.com/6947168/98615826-1b7ad780-22c1-11eb-869f-ea0b2daf8091.png)

PR #639, opened 2020-11-06
Slack Notification Plugin - add field Message Backlog Limit.

lingpri:
**Test evidence with a backlog message override set to 7**

backlog_size

**Slack Notification - Message Backlog Override.**

backlog_messages


src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -154,27 +159,38 @@ private String buildMessageTitle(EventNotificationContext ctx) {
...
 	String buildCustomMessage(EventNotificationContext ctx, SlackEventNotificationConfig config, String template) throws PermanentEventNotificationException {
 		List<MessageSummary> backlog = getAlarmBacklog(ctx);
-		Map<String, Object> model = getCustomMessageModel(ctx, config, backlog);
-		try {
-			LOG.debug("template = {} model = {}" ,template, model);
-			return templateEngine.transform(template, model);
-		} catch (Exception e) {
-			LOG.error("Exception during templating [{}]", e.toString());
-			throw new PermanentEventNotificationException(e.toString(),e.getCause());
+		backlog = getMessageBacklog(config, backlog);
```
bernd, line 162: Overwriting existing variables can be quite confusing and can easily lead to errors. Please create a new variable to store the modified backlog. (e.g. `truncatedBacklog`)

BTW, that's one of the reasons why I usually mark every variable and field as `final` if possible. That way you can't overwrite existing variables by accident.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -154,27 +159,38 @@ private String buildMessageTitle(EventNotificationContext ctx) {
...
+
+		Map<String, Object> model = getCustomMessageModel(ctx, config.type(), backlog);
+			try {
+				LOG.debug("template = {} model = {}", template, model);
+				return templateEngine.transform(template, model);
+			} catch (Exception e) {
+				LOG.error("Exception during templating [{}]", e.toString());
+				throw new PermanentEventNotificationException(e.toString(), e.getCause());
```
bernd, line 170: The indentation here seems to have changed and doesn't conform to the default IntelliJ indentation. What's the reason for this?

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -154,27 +159,38 @@ private String buildMessageTitle(EventNotificationContext ctx) {
...
+		}
+
+
+	}
+
+	List<MessageSummary> getMessageBacklog(SlackEventNotificationConfig config, List<MessageSummary> backlog) {
+		if(config.backlogSize() > 0 && backlog != null) {
+			backlog = backlog.stream().limit(config.backlogSize()).collect(Collectors.toList());
```
bernd, line 178: Similar to my comment above, modifying parameters can be confusing and can easily lead to errors. Please return a new list instead of modifying the parameter that gets passed in.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -154,27 +159,38 @@ private String buildMessageTitle(EventNotificationContext ctx) {
...
+				return templateEngine.transform(template, model);
+			} catch (Exception e) {
+				LOG.error("Exception during templating [{}]", e.toString());
+				throw new PermanentEventNotificationException(e.toString(), e.getCause());
+		}
+
+
+	}
+
+	List<MessageSummary> getMessageBacklog(SlackEventNotificationConfig config, List<MessageSummary> backlog) {
```
bernd, line 176: We usually annotate package-private methods with a `@VisibleForTesting` annotation to make other developers know why the method isn't private.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -154,27 +159,38 @@ private String buildMessageTitle(EventNotificationContext ctx) {
...
+		return backlog;
 	}
 
 
 	List<MessageSummary> getAlarmBacklog(EventNotificationContext ctx) {
 		return notificationCallbackService.getBacklogForEvent(ctx);
 	}
 
-	Map<String, Object> getCustomMessageModel(EventNotificationContext ctx, SlackEventNotificationConfig config, List<MessageSummary> backlog) {
+	Map<String, Object> getCustomMessageModel(EventNotificationContext ctx, String type, List<MessageSummary> backlog) {
```
bernd, line 188: What's the reason for changing the parameter from `SlackEventNotificationConfig` to `String`? This doesn't seem to be related to the backlog change.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -41,9 +42,15 @@
 @JsonDeserialize(builder = SlackEventNotificationConfig.Builder.class)
 public abstract class SlackEventNotificationConfig implements EventNotificationConfig {
 
-	private final String regex = "https:\\/\\/hooks.slack.com\\/services\\/";
-	private final Pattern pattern = Pattern.compile(regex, Pattern.MULTILINE);
-	public static final String TYPE_NAME = "slack-notification-v1";
+	private final  String regex = "https:\\/\\/hooks.slack.com\\/services\\/";
+	private final  Pattern pattern = Pattern.compile(regex, Pattern.MULTILINE);
+	public static  final String TYPE_NAME = "slack-notification-v1";
```
bernd, line 47: The indentation changes here are unrelated. What's the reason for them? We usually just use IntelliJ's standard formatting. (we will soon have our custom style config incl. auto-formatting)

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -106,12 +119,16 @@ public ValidationResult validate() {
 		ValidationResult validation =  new ValidationResult();
 		final Matcher matcher = pattern.matcher(webhookUrl());
 
+		if(backlogSize() < 0 || backlogSize() > MAX_BACKLOG_MESSAGES) {
```
bernd, line 122: Why do we impose a maximum number of messages for the backlog size?

src/main/java/org/graylog/integrations/notifications/types/SlackMessage.java
```diff
@@ -73,6 +76,7 @@ public String getJsonString() {
 		params.put("channel", channel);
 		params.put("text", message);
 		params.put("link_names", linkNames);
+		params.put("backlog_size",backlogSize);
```
bernd, line 79: Why do you add the `backlog_size` to the payload that we send to Slack? Is this an officially supported field?

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -100,7 +97,8 @@ private NotificationDto getHttpNotification() {
 
     @Test
     public void createSlackMessage() throws EventNotificationException {
-       String expected = "{\"link_names\":true,\"attachments\":[{\"fallback\":\"Custom Message\",\"text\":\"a custom message\",\"pretext\":\"Custom Message:\",\"color\":\"#FF2052\"}],\"channel\":\"#general\",\"text\":\"@channel *Alert _Event Definition Test Title_* triggered:\\n> Event Definition Test Description \\n\"}";
+       //String expected = "{\"link_names\":true,\"attachments\":[{\"fallback\":\"Custom Message\",\"text\":\"a custom message\",\"pretext\":\"Custom Message:\",\"color\":\"#FF2052\"}],\"channel\":\"#general\",\"text\":\"@channel *Alert _Event Definition Test Title_* triggered:\\n> Event Definition Test Description \\n\",\"backlogSize\":1}";
```
bernd, line 100: Can this commented code be removed?

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import cloneDeep from 'lodash/cloneDeep';
 import get from 'lodash/get';
 
-import { Input } from 'components/bootstrap';
 import { getValueFromInput } from 'util/FormsUtils';
+import { Input } from 'components/bootstrap';
```
bernd, line 7: What's the reason for this change?

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -77,6 +79,7 @@ class SlackNotificationForm extends React.Component {
       link_names: false,
       icon_url: '',
       icon_emoji: '',
+      backlog_size: 5,
```
bernd, line 82: In the backend the default `backlogSize` is 0. I think we should use the same in the backend and frontend.

bernd (changes requested):

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: nick gave a suggestion to default the value to `5' I have changed the backend to default it to 5 as well.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: i had to revert a few changes in my front end code related to the task enabling and disabling of the backlog message field. probably, it changed the order of imports.

src/main/java/org/graylog/integrations/notifications/types/SlackMessage.java
lingpri: we are overriding the EventNotificationSettings.FIELD_BACKLOG_SIZE using this field backlog_size and yes its an officially supported field in the graylog core code hence the override.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
lingpri: just so we don't overload the slack client with messages, that it can't handle. we can tune the paramter, as we do more performance testing.

src/main/java/org/graylog/integrations/notifications/types/SlackMessage.java
lingpri: Good point, we don't need this.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: its reverted back to 0  in this [commit]( https://github.com/Graylog2/graylog-plugin-integrations/pull/639/commits/2f2efcfef97a5b3d7ce78a0866a93148bb4eb5ff)

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -144,6 +175,27 @@ class SlackNotificationForm extends React.Component {
...
+              </InputGroup.Addon>
+              <FormControl type="number"
+                           id="backlog_size"
+                           name="backlog_size"
+                           onChange={this.handleBacklogSizeChange}
+                           value={backlogSize}
+                           min="0"
+                           disabled={!isBacklogSizeEnabled} />
+            </InputGroup>
+            <HelpBlock>Limit the number of backlog messages that will be included in Slack notifications.</HelpBlock>
```
bernd, line 196: If we have a maximum backlog size, we should also note that in the help block. Otherwise users might enter a higher number than the maximum and only will only be aware of it after the validation runs.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -35,148 +36,161 @@
...
+        return backlog;
+    }
+
+    @VisibleForTesting
+    List<MessageSummary> getAlarmBacklog(EventNotificationContext ctx) {
+        return notificationCallbackService.getBacklogForEvent(ctx);
+    }
+
+    @VisibleForTesting
+    Map<String, Object> getCustomMessageModel(EventNotificationContext ctx, String type, List<MessageSummary> backlog) {
```
bernd, line 181: What is the reason changing the second argument form `SlackEventNotificationConfig` to `String`? That doesn't seem to be related to the backlog change.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -35,148 +36,161 @@
...
+        }
+    }
+
+    @VisibleForTesting
+    List<MessageSummary> getMessageBacklog(SlackEventNotificationConfig config, List<MessageSummary> backlog) {
+        List<MessageSummary> truncatedBacklog = Collections.emptyList();
+        if (config.backlogSize() > 0 && backlog != null) {
+            truncatedBacklog = backlog.stream().limit(config.backlogSize()).collect(Collectors.toList());
+            return truncatedBacklog;
+        }
```
bernd, line 171: No need to create a new variable here. You can just return the new one. Also, `Collections.emptyList()` returns an immutable list and thus shouldn't be used in contexts where it might be modified.

```suggestion
        if (config.backlogSize() > 0 && backlog != null) {
            return backlog.stream().limit(config.backlogSize()).collect(Collectors.toList());
        }
```

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -41,140 +42,165 @@
...
-	static final String FIELD_CHANNEL = "channel";
-	static final String FIELD_CUSTOM_MESSAGE = "custom_message";
-	static final String FIELD_USER_NAME = "user_name";
-	static final String FIELD_NOTIFY_CHANNEL = "notify_channel";
-	static final String FIELD_LINK_NAMES = "link_names";
-	static final String FIELD_ICON_URL = "icon_url";
-	static final String FIELD_ICON_EMOJI = "icon_emoji";
+    private final String regex = "https:\\/\\/hooks.slack.com\\/services\\/";
+    private final Pattern pattern = Pattern.compile(regex, Pattern.MULTILINE);
+    private static final long MAX_BACKLOG_MESSAGES = 49;
```
bernd, line 49: Why are we using a default of `49` instead of like `50`? That seems odd.

bernd (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
lingpri: the junit tests were becoming simple, with a string datatype.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
lingpri: This number is subject to change, when we discover new facts . I will make it 50 for now. Yeah 49 is odd :)

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -35,148 +36,160 @@
...
+    public void execute(EventNotificationContext ctx) throws PermanentEventNotificationException, TemporaryEventNotificationException {
+        final SlackEventNotificationConfig config = (SlackEventNotificationConfig) ctx.notificationConfig();
+        ValidationResult result = config.validate();
+        result.getErrors().entrySet().stream().forEach(e -> LOG.error("Invalid configuration for key [{}] and value [{}]", e.getKey(), e.getValue()));
+
+        LOG.debug("SlackEventNotification backlog size in method execute is [{}]", config.backlogSize());
+
+        if (result.failed()) {
+            throw new PermanentEventNotificationException("Please verify your Slack Event Configuration");
+        }
```
waab76, line 87: As mentioned previously, I don't think there's any reason to call config.validate() here. The validation happens when a user creates the new Notification via the UI. This should be removed.
```suggestion
        LOG.debug("SlackEventNotification backlog size in method execute is [{}]", config.backlogSize());
```

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -35,148 +36,160 @@
...
+    }
+
+    private String buildMessageTitle(EventNotificationContext ctx) {
+        String eventDefinitionName = ctx.eventDefinition().map(EventDefinitionDto::title).orElse("Unnamed");
+        return "_" + eventDefinitionName + "_";
+    }
+
+    String buildCustomMessage(EventNotificationContext ctx, SlackEventNotificationConfig config, String template) throws PermanentEventNotificationException {
+        final List<MessageSummary> backlog = getAlarmBacklog(ctx);
+        final List<MessageSummary> truncatedBacklog = getMessageBacklog(config, backlog);
```
waab76, line 153: Why have a separate method for truncating the backlog?  Wouldn't it be simpler to pass the config or a limit into the `getAlarmBacklog` method and just get the appropriate backlog returned?

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -1,16 +1,16 @@
...
+ * <p>
  * Graylog is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- *
+ * <p>
  * You should have received a copy of the GNU General Public License
  * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
  */
```
waab76, line 16: Again wondering why the change to the license info

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -41,140 +42,165 @@
...
+
+    @Override
+    @JsonIgnore
+    public ValidationResult validate() {
+        ValidationResult validation = new ValidationResult();
+        final Matcher matcher = pattern.matcher(webhookUrl());
+
+        if (backlogSize() < 0 || backlogSize() > MAX_BACKLOG_MESSAGES) {
+            validation.addError(FIELD_BACKLOG_SIZE, INVALID_BACKLOG_ERROR_MESSAGE);
+        }
```
waab76, line 130: When `backlogSize` is set to zero, we do not limit the backlog at all.  Thus, having an upper limit on `backlogSize` makes no sense.  If the user has decided to set the backlog for their Alert to 300 and we will allow them to get all 300, why would we not allow them to get 100 or 200?

src/main/java/org/graylog/integrations/notifications/types/SlackMessage.java
```diff
@@ -1,16 +1,16 @@
...
+ * <p>
  * Graylog is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- *
+ * <p>
  * You should have received a copy of the GNU General Public License
  * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
  */
```
waab76, line 16: Changing the license?

src/test/java/org/graylog/integrations/notifications/types/SlackClientTest.java
```diff
@@ -54,7 +54,7 @@ public void send_sendsHttpRequestAsExpected_whenInputIsGood() throws Exception {
         assertThat(sent.url().toString()).isEqualTo("http://url.com/");
         assertThat(sent.method()).isEqualTo("POST");
         assertThat(sent.body()).isNotNull();
-        assertThat(sent.body().contentLength()).isEqualTo(76);
+        assertThat(sent.body().contentLength()).isEqualTo(76L);
```
waab76, line 57: You can use `okio.Buffer` to check that the message sent in the request matches the message you expect (also, it might help to use a String constant or String variable rather than a String literal to make comparison and size checks easier).
```suggestion
        assertThat(sent.body().contentLength()).isEqualTo(Long.valueof(TEST_MESSAGE.length()));
        Buffer buffer = new Buffer();
        sent.body().writeTo(buffer);
        assertThat(buffer.readUtf8()).isEqualTo(TEST_MESSAGE);
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfigTest.java
```diff
@@ -30,4 +32,39 @@ public void validate_failsAndReturnsAnError_whenWebhookUrlIsInvalid() {
         Map errors = result.getErrors();
         assertThat(errors).size().isGreaterThan(0);
     }
+
+    @Test
+    public void validate_messageBacklog() {
```
waab76, line 37: Generally speaking, a test case should only be testing one behavior at a time.  It would probably be good to split this into three separate tests for the three different conditions being tested.  

As mentioned earlier in the PR, I don't think that we should be enforcing an upper limit on `backlogSize` so that eliminates one test.  However, since zero is intended to have special meaning, you should have a test verifying that zero is a valid value.

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -145,26 +141,78 @@ public void getCustomMessageModel() {
...
+
+    @Test
+    public void test_customMessage_With_Message_Backlog_Override() throws PermanentEventNotificationException {
+
+        SlackEventNotificationConfig slackConfig = SlackEventNotificationConfig.builder()
+                .backlogSize(5)
+                .build();
+        String message = slackEventNotification.buildCustomMessage(eventNotificationContext, slackConfig, "Ich spreche Deutsch");
+        assertThat(message).isEqualTo("Ich spreche Deutsch");
+    }
```
waab76, line 175: What is this actually testing?  How is the behavior you're validating related to the backlog size override?

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -145,26 +141,78 @@ public void getCustomMessageModel() {
...
+
+
+    @Test
+    public void test_backlog_messagesize() {
```
waab76, line 179: Again, you've got multiple test cases rolled into a single test method.  These probably deserve to be their own individual tests.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -141,9 +174,31 @@ class SlackNotificationForm extends React.Component {
                  label="Custom Message (optional)"
                  type="textarea"
                  bsStyle={validation.errors.custom_message ? 'error' : null}
-                 help={get(validation, 'errors.custom_message[0]', 'Custom message to be appended below the alert title. See https://docs.graylog.org/en/latest/pages/alerts.html#data-available-to-notifications for more details.')}
+                 help={get(validation, 'errors.custom_message[0]', element)}
                  value={config.custom_message || ''}
                  onChange={this.handleChange} />
+
+          <FormGroup>
+            <ControlLabel>Backlog Message Items (optional)</ControlLabel>
```
waab76, line 182: You used `Message Backlog Limit` on the Details page.  I think you should use the same here.
```suggestion
            <ControlLabel>Message Backlog Limit (optional)</ControlLabel>
```

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -141,9 +174,31 @@ class SlackNotificationForm extends React.Component {
...
+                       checked={isBacklogSizeEnabled}
+                       onChange={this.toggleBacklogSize} />
+              </InputGroup.Addon>
+              <FormControl type="number"
+                           id="backlog_size"
+                           name="backlog_size"
+                           onChange={this.handleBacklogSizeChange}
+                           value={backlogSize}
+                           min="0"
+                           max="50"
```
waab76, line 196: Don't think we need a max value.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -141,9 +174,31 @@ class SlackNotificationForm extends React.Component {
...
+                           disabled={!isBacklogSizeEnabled} />
+            </InputGroup>
+            <HelpBlock>Slack notification cannot have more than 50 backlog messages.</HelpBlock>
```
waab76, line 199: This does not do a very good job of explaining what this field is for.  Also, I don't think we should be putting an upper limit on the value.  This `HelpBlock` should describe what this field is for and also make it clear what will happen when set to a special value (zero).

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
```diff
@@ -1,52 +1,55 @@
...
+      <tr>
+        <td>Channel</td>
+        <td>{notification.config.channel}</td>
+      </tr>
+      <tr>
+        <td>Custom Message</td>
+        <td>{notification.config.custom_message}</td>
+      </tr>
+      <tr>
+        <td>Backlog Message Size</td>
```
waab76, line 25: Again, we should be consistent with the labeling for this value across all three pages.

waab76 (changes requested):

src/web/event-notifications/event-notification-details/SlackNotificationDetails.jsx
```diff
@@ -23,6 +23,7 @@ const SlackNotificationDetails = ({ notification }) => (
                            {notification.config.custom_message || <em>Empty body</em>}
                          </NewExampleWell>
                        )} />
+    <ReadOnlyFormGroup label="Message Backlog Limit" value={notification.config.backlog_size} />
```
waab76, line 26: Should we be displaying something like `No limit` or `Disabled` when `backlog_size` is zero?

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -141,9 +174,31 @@ class SlackNotificationForm extends React.Component {
...
+                       type="checkbox"
+                       checked={isBacklogSizeEnabled}
+                       onChange={this.toggleBacklogSize} />
+              </InputGroup.Addon>
+              <FormControl type="number"
+                           id="backlog_size"
+                           name="backlog_size"
+                           onChange={this.handleBacklogSizeChange}
+                           value={backlogSize}
+                           min="0"
```
waab76, line 195: Since users seem to be able to turn this on and off with a checkbox, I am concerned about the behavior when they turn it on and set the value to zero.  

As a user, if I turn on the limit and set the limit to 0, I would expect to see my backlog messages limited to zero rather than unlimited.

Perhaps we should be using -1 as the flag value rather than zero?  Or we should make the minimum (when active) 1?  It just seems like we are creating a potentially confusing user experience.

waab76 (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
lingpri: hope, this is a minor issue and can be fixed with a simple mvn command.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
lingpri: the slack client throws Http_error_code 500 , when the message goes beyond a certain limit. I would like to impose a limit, on the number of messages being sent to the slack client, just so we don't break the Slack client. Its a good practice to set the upper limit. The number 50 is empirical and is subject to change as we find new facts.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
waab76: But why did you decide to change the license as part of this PR?

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: the docs has more verbose explanation.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: message_backlog

The parent field has this field name. I think we should sync up with this name and description to be consistent.let me know your thoughts.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: its being more defensive, just so we don't accidentally flood the slack client . Let me know your thoughts.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: please refer screen shot above, just so we are consistent, with the parent field name and description.

src/web/event-notifications/event-notification-types/SlackNotificationSummary.jsx
lingpri: yeah agree. the original story uses multiple conventions and they both mean the same,

https://github.com/Graylog2/graylog-plugin-integrations/issues/623

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: the behavior is well documented in `test_backlog_messagesize`method and applying this change, will lead to a breaking unit tests. if you have further concerns, please create a bug and we can address this in a seperate issue. 

 //global backlog messages is 0 ,when the SlackEventNotificationConfig backlog_size is set to 0, then the expected backlog message size is 0.

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
lingpri: the sut is `getMessageBacklog` and it covers all the scenarios of the `if` condition. is this a blocking issue?

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
lingpri: i didn't intend to change. I noticed this, only after I raised the PR.

src/web/event-notifications/event-notification-details/SlackNotificationDetails.jsx
lingpri: sorry i donot follow your question .

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
lingpri: Thanks for your feedback. I will perform some more testing locally, before I commit this suggestion.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
lingpri: please refer to bernd's earlier comments, he prefers a `immutable` list

src/web/event-notifications/event-notification-details/SlackNotificationDetails.jsx
waab76: Non-issue.  Disregard.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
waab76: Comments in a JUnit are not the same as user-facing documentation.  Here is what I see when I test this locally:

* I open the configuration for my Slack notification
* I check the box for message backlog limit
* I set the value to 0 - as a User, I expect that I will receive 0 backlog messages in my notification
* I save my changes
* When the next notification happens, I see unlimited backlog messages in the notification
* I re-open the configuration for my Slack notification
* I see that the message backlog limit checkbox is unchecked

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
waab76: And since you have control over what will be returned by `getAlarmBacklog`, that method can return an immutable list if desired.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
waab76: My biggest point here is that it makes no sense to put a max on the backlog limit because the backlog limit can be disabled and the entire backlog will flow through.

The amount of data that gets POSTed to the Slack webhook is entirely dependent on the message template being used.  If the user is dumping the full contents of each backlog message, then 5 messages may be enough to produce an oversized Slack message.  If the user is only including message IDs, then 500 messages might be fine.  But if the user sets their message backlog to N and they can cause all N backlog messages to be sent by disabling the limit, why should we not allow them to set their limit to values from 51 to (N - 1)?

A quick Google search pointed to the [Slack documentation](https://api.slack.com/changelog/2018-04-truncating-really-long-messages), which indicates that any message with more than 40000 characters of text will be truncated on the server side.  I have validated this locally with a 60000-character message template (you can see some of the truncated messages in our test Slack channel).  I should note that these auto-truncated messages returned HTTP 200 when POSTed to the webhook.

Given that Slack will auto-truncate overly large messages, I don't think there's a way for us to come up with a correct upper limit.  It would make more sense (knowing about Slack's auto-truncate behavior) for us to shorten the message on our side before sending it to Slack and put a warning in the logs than to put an arbitrary cap on

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
waab76: My thoughts have been included above.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
lingpri: I will remove the max_backlog_messages as per your advice.

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
lingpri: if i understand right, can you please rewrite your thoughts like below, just so we stay consistent in our language, 
 
//global backlog message is 10,when the SlackEventNotificationConfig backlog_size is set to 0, then the expected backlog message size is 10.

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
waab76: Not a blocking issue, just general informational feedback about good practices for writing unit tests.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -41,140 +42,165 @@
...
+    private final String regex = "https:\\/\\/hooks.slack.com\\/services\\/";
+    private final Pattern pattern = Pattern.compile(regex, Pattern.MULTILINE);
+    private static final long MAX_BACKLOG_MESSAGES = 50;
+    private static final String HEX_COLOR = "#ff0500";
 
+    private static final String CUSTOM_MESSAGE = "hello World";
+    private static final boolean FALSE = false;
+    private static final long BACKLOG_SIZE = 0;
+
+    static final String INVALID_BACKLOG_ERROR_MESSAGE = "Backlog size cannot be less than zero or greater than " + MAX_BACKLOG_MESSAGES + ".";
```
waab76, line 56: Now that the max is not being enforced, this message needs to be cleaned up.  This is a blocker.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
waab76: Not a blocker, but I'm going to keep bringing it up every time this code is up for review.

waab76 (changes requested):
Thanks for all the changes so far.  I think the only remaining blocker is the error message I called out in SlackNotificationConfig.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -125,10 +122,6 @@ public ValidationResult validate() {
         ValidationResult validation = new ValidationResult();
         final Matcher matcher = pattern.matcher(webhookUrl());
 
-        if (backlogSize() < 0 ) {
-            validation.addError(FIELD_BACKLOG_SIZE, INVALID_BACKLOG_ERROR_MESSAGE);
-        }
```
waab76: This check should remain.  Without it, users can send a test notification with `backlog_size` of -1.

waab76 (changes requested):

waab76 (approved):
All blockers have been addressed.  Please wait for a green build and approval from @bernd before merging to `master`

src/web/event-notifications/event-notification-types/SlackNotificationForm.jsx
```diff
@@ -157,9 +190,30 @@ class SlackNotificationForm extends React.Component {
...
+                           disabled={!isBacklogSizeEnabled} />
+            </InputGroup>
+            <HelpBlock>Limit the number of backlog messages sent as part of the Slack notification.</HelpBlock>
```
waab76, line 214:
```suggestion
            <HelpBlock>Limit the number of backlog messages sent as part of the Slack notification.  If set to 0, no limit will be enforced.</HelpBlock>
```

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
```diff
@@ -161,26 +152,83 @@ public void getCustomMessageModel() {
     }
 
 
-    @Test (expected = PermanentEventNotificationException.class)
+    @Test
+    @Ignore("Disabled, unable to test this method, since it return void")
```
waab76, line 156: This does not make sense.  There are plenty of ways to meaningfully test a method that returns void.  Unit testing is not just about ensuring a method returns the expected value.  For methods like `execute`, whose job is to coordinate the action, you can validate that the right dependencies are being invoked with the right parameters.  You can validate that the error handling goes according to plan ("when dependency X throws a NPE, my method should throw a PermanentEventNotificationException", things like that).

If you need examples, feel free to look at [PagerDutyNotifiationTest](https://github.com/Graylog2/graylog-plugin-integrations/blob/master/src/test/java/org/graylog/integrations/pagerduty/PagerDutyNotificationTest.java)

The only time we should be removing unit tests is when we're removing the functionality they tested.

waab76 (changes requested):
Please update the help text on the new config field to make it clear to users that a value of 0 will result in no limit being applied.  Also, please reconsider the decision to abandon unit testing on `SlackEventNotification.execute`

src/test/java/org/graylog/integrations/notifications/types/SlackEventNotificationTest.java
lingpri: Thanks for the example.

waab76 (approved):
I believe all my blockers have been addressed.  Local testing works as expected.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +38,143 @@
...
+
+            LOG.error("SlackEventNotification send error for id {} : {}", ctx.notificationId(), e.toString());
+            final Notification systemNotification = notificationService.buildNow()
+                    .addNode(nodeId.toString())
+                    .addType(Notification.Type.GENERIC)
+                    .addSeverity(Notification.Severity.NORMAL)
+                    .addDetail("SlackEventNotification send error ", e.toString());
+
+            notificationService.publishIfFirst(systemNotification);
+            throw new EventNotificationException("Slack notification is triggered, but sending failed. " + e.getMessage(), e);
```
bernd, line 92: This exception handling block is catching `Exception`. Since the `SlackClient#send` method is already returning a `TemporaryEventNotificationException` or `PermanentEventNotificationException`, we need a separate exception handling block that catches `EventNotificationException`s.

Otherwise we wrap the `EventNotificationException` returned from `SlackClient#send` in another `EventNotificationException` here and the scheduler will never retry a `TemporaryEventNotificationException`. That means notifications can be lost.

bernd (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+        } catch (PermanentEventNotificationException e) {
+
+            LOG.error("SlackEventNotification send error for id {} : {}", ctx.notificationId(), e.toString());
+            final Notification systemNotification = notificationService.buildNow()
+                    .addNode(nodeId.toString())
+                    .addType(Notification.Type.GENERIC)
+                    .addSeverity(Notification.Severity.URGENT)
+                    .addDetail("SlackEventNotification send error ", e.toString());
+            notificationService.publishIfFirst(systemNotification);
+            throw new TemporaryEventNotificationException(String.format("Error sending the SlackNotification message :: %s", e.getMessage()), e);
```
bernd, line 97: This is wrong. If you catch a permanent error you shouldn't re-throw a temporary one. Otherwise the scheduler will retry a permanent exception over and over again.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+
+        } catch (Exception exp) {
+            LOG.error("SlackEventNotification send error for id {} : {}", ctx.notificationId(), exp.toString());
+            throw exp;
```
bernd, line 101: Here we must wrap the caught exception in a generic `EventNotificationException` to make sure it's properly handled by the scheduler.

bernd (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+                    .addNode(nodeId.toString())
+                    .addType(Notification.Type.GENERIC)
+                    .addSeverity(Notification.Severity.URGENT)
+                    .addDetail("SlackEventNotification send error ", e.toString());
+            notificationService.publishIfFirst(systemNotification);
+            throw new PermanentEventNotificationException(String.format("Error sending the SlackNotification message :: %s", e.getMessage()), e);
+
+        } catch (Exception exp) {
+            LOG.error("SlackEventNotification send error for id {} : {}", ctx.notificationId(), exp.toString());
+            throw new EventNotificationException("There was an exception triggering the PagerDuty event.", exp);
```
bernd, line 101: The message talks about a "PagerDuty" event.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+        } catch (PermanentEventNotificationException e) {
+
+            LOG.error("SlackEventNotification send error for id {} : {}", ctx.notificationId(), e.toString());
+            final Notification systemNotification = notificationService.buildNow()
+                    .addNode(nodeId.toString())
+                    .addType(Notification.Type.GENERIC)
+                    .addSeverity(Notification.Severity.URGENT)
+                    .addDetail("SlackEventNotification send error ", e.toString());
+            notificationService.publishIfFirst(systemNotification);
+            throw new PermanentEventNotificationException(String.format("Error sending the SlackNotification message :: %s", e.getMessage()), e);
```
bernd, line 97: Since we already get a `PermanentEventNotificationException`, we can just re-throw it without wrapping it into another `PermanentEventNotificationException`.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+        LOG.debug("SlackEventNotification backlog size in method execute is [{}]", config.backlogSize());
+
+        try {
+            SlackMessage slackMessage = createSlackMessage(ctx, config);
+            slackClient.send(slackMessage, config.webhookUrl());
+        } catch (TemporaryEventNotificationException exp) {
+            //scheduler needs to retry a TemporaryEventNotificationException
+            LOG.error("SlackEventNotification send error for id {} : {}", ctx.notificationId(), exp.toString());
+
+            throw new TemporaryEventNotificationException(String.format("Error sending the SlackNotification message :: %s", exp.getMessage()), exp);
```
bernd, line 87: Since we already get a `TemporaryEventNotificationException`, we can just re-throw it without wrapping it into another `TemporaryEventNotificationException`.

bernd (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+
+        LOG.debug("SlackEventNotification backlog size in method execute is [{}]", config.backlogSize());
+
+
+        try {
+            SlackMessage slackMessage = createSlackMessage(ctx, config);
+            slackClient.send(slackMessage, config.webhookUrl());
+        } catch (TemporaryEventNotificationException exp) {
+            //scheduler needs to retry a TemporaryEventNotificationException
+            LOG.error(errorMessage, exp.toString());
```
bernd, line 88: The `errorMessage` message template doesn't include a `{}` placeholder for the argument you are passing to `LOG.error`. That means the logged error message will not include the argument value.

When you look at the exception handler in `EventNotificationExecutionJob` you will see that it already includes the notification ID and title in the wrapping `JobExecutionException`. (which will then be logged by the `JobExecutionEngine`. That means you don't really need to log this in the exception handling code in this method.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+            throw exp;
+        } catch (PermanentEventNotificationException e) {
+            LOG.error(errorMessage, e.toString());
```
bernd, line 91: Same as above.

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
```diff
@@ -38,145 +39,150 @@
...
+            final Notification systemNotification = notificationService.buildNow()
+                    .addNode(nodeId.toString())
+                    .addType(Notification.Type.GENERIC)
+                    .addSeverity(Notification.Severity.URGENT)
+                    .addDetail(errorMessage, e.toString());
+            notificationService.publishIfFirst(systemNotification);
+            throw e;
+
+        } catch (Exception exp) {
+            LOG.error(errorMessage, exp.toString());
```
bernd, line 101: Same here.

bernd (changes requested):

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java
lingpri: the argument value has been appended preemptively like below and hence doesn't need a `{}` placeholder.
`final String errorMessage = "SlackEventNotification execute error for id : " + ctx.notificationId();`

bernd: But then you don't need to pass `exp.toString()` because it will just be ignored.

lingpri: i was trying to repurpose the errormessage across `log.error` and while constructing the exception and missed the idea of Log.error 2nd argment is a filler for the 1st argument.

bernd (approved):
Thanks for the fixes!

PR #640, opened 2020-11-06
Use Icon component from 3.3 core

kyleknighted:
We were using obsolete icons - updated to component from Core


linuspahl (approved):
Looks good to me.

PR #657, opened 2020-11-18
Disable cloudwatch on cloud

Gaya:
We can't use CloudWatch inputs right now on Cloud so we need to hide it.

### DIFF
--- a/src/web/index.jsx
+++ b/src/web/index.jsx
@@ -17,6 +17,7 @@
 import 'webpack-entry';
 
 import { PluginManifest, PluginStore } from 'graylog-web-plugin/plugin';
+import AppConfig from 'util/AppConfig';
 
 import Routes from 'aws/common/Routes';
 
@@ -61,4 +62,6 @@ const manifest = new PluginManifest(packageJson, {
   ],
 });
 
-PluginStore.register(manifest);
+if (!AppConfig.isCloud()) {
+  PluginStore.register(manifest);
+}


Gaya:
See https://github.com/Graylog2/graylog-plugin-cloud/issues/99

Gaya:
Need to update the `cloud-4.0` manifest to use `cloud-4.0` first. Will make separate PR for that

Gaya:
See https://github.com/Graylog2/graylog-project-internal/pull/51

danotorrey:
On the Input Profile Details page > Create Input page, the AWS Kinesis/Cloudwatch input still shows up. This is because the list of inputs to add come from the backend (VIA `GET /plugins/org.graylog.plugins.cloud/forwarder/inputs/types`)
image

I am not sure at the moment how to hide specific inputs on the backed. I am not sure yet how to work around the issue mentioned in `https://github.com/Graylog2/graylog-project-internal/pull/51#issuecomment-735839943` yet. I will give it some thought.

Gaya:
@danotorrey ah, I didn't know it was possible to add this input this way too.
What I can do is filter it out on the frontend, but my guess it's better to not register the input on the backend. Let me know if I can help

edmundoa:
@danotorrey in the team meeting yesterday I understood that the approach we want to take is to not bind the message Input, which would also mean the Input type is not available. As far as I can see, making this line to execute on not-cloud environments would suffice to fix this particular case:

https://github.com/Graylog2/graylog-plugin-integrations/blob/2928e3bdd823d5ae29fdbafd4d9c44ee15edf7a9/src/main/java/org/graylog/integrations/IntegrationsModule.java#L128

Do you see any concerns in doing that? Filtering the Input (in the back-end or UI) would also be an option but I think there will be more chances for something to go wrong.

danotorrey:
> Do you see any concerns in doing that? Filtering the Input (in the back-end or UI) would also be an option but I think there will be more chances for something to go wrong.

@edmundoa I really like this backend-filtering approach, but the problem is that in the `IntegrationsModule. configure()` method, I do not think we are able to read/inject application configuration properties from the server.conf file (in order to read the `is_cloud` flag) at this point in the server startup, because it is executed really early in the server startup process, and the Graylog [configuration file parsing is still being processed](https://github.com/Graylog2/graylog2-server/blob/29aa0ab4bd8dca7ef60ab7b5f1b17b805d9e7027/graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java#L277).

@edmundoa @bernd @thll Do you have any ideas on a workaround? For skipping inputs for the Forwarder, we used [a system property as a workaround](https://github.com/Graylog2/graylog-plugin-enterprise-integrations/blob/d5c9e647bc8764406da00db9380d539e9d67a77a/src/main/java/org/graylog/enterprise/integrations/EnterpriseIntegrationsModule.java#L160-L168), but not sure I see that being useful here. 

Another idea is to add a constant list of the names/types of ignored inputs in the [ForwarderInputTypesResource](https://github.com/Graylog2/graylog-plugin-cloud/blob/309abfa1b7ea05c6df58f274d40c462782e53577/server-plugin/src/main/java/org/graylog/plugins/cloud/forwarder/rest/resources/ForwarderInputTypesResource.java#L44) and just filter them out before returning the list of available inputs to the UI.

bernd:
@danotorrey As you said, the modules don't have access to the configuration flag at that point. This can be fixed, of course. But it would require a larger refactoring of the plugin system to be able to pass the configuration to the modules.

I think introducing a property like `graylog.cloud` would be a pragmatic way to solve this issue. We can add a `#isCloud` method to the `PluginModule` class so we don't have to repeat the property check in different plugins.

Having a block or allow list of inputs for cloud would only work for inputs. But I think we also need this check for other things like lookup tables. (e.g. we might want to exclude data adapters that don't work in cloud, like the CSV file one)

danotorrey:
> I think introducing a property like `graylog.cloud` would be a pragmatic way to solve this issue. We can add a `#isCloud` method to the `PluginModule` class so we don't have to repeat the property check in different plugins.

@bernd Sounds great. Thanks for the suggestion. I will do this.

danotorrey:
@Gaya Since we've found a way to address this in the backed, I think we can close this PR #657 and https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/425. I'll open new PRs for hiding the inputs on the backend.

I just merged the manifest changes PR https://github.com/Graylog2/graylog-project-internal/pull/51.

Gaya:
> @Gaya Since we've found a way to address this in the backed, I think we can close this PR #657 and [Graylog2/graylog-plugin-enterprise-integrations#425](https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/425). I'll open new PRs for hiding the inputs on the backend.
> 
> I just merged the manifest changes PR [Graylog2/graylog-project-internal#51](https://github.com/Graylog2/graylog-project-internal/pull/51).

Thanks, Dan. Will do!

PR #658, opened 2020-11-19
Palo Alto input updates to support PANOS v10 schemas

waab76:
Palo Alto Networks has released the schemas for PAN-OS 10.  This change updates the Palo Alto 9.x input to support the PAN-OS 10 schemas and also makes some other requested field changes.  One additional change in this PR is that when multiple instances of the same key are used to build a Palo Alto message, the message field will be converted to a list of all unique values.

The changes to existing fields and list conversion behavior were tested with JUnit tests.  No test data is available to test the PAN-OS 10 schema changes.


bud1979 (approved):
Looks good from my notes.

waab76:
Still trying to figure out why this JUnit test is failing.  It's not failing locally, which is making things tough.

PR #686, opened 2020-12-22
Update Guice config to split server-only and universal

waab76:
Forwarder was failing to start with current Guice config because it included Notifications that depend on services not present in the Forwarder.  This change splits the Guice config into `server` and `universal` sections and moves the Notifications into the `server` section.

Verified that the Forwarder successfully starts with the modified config.


danotorrey (approved):
LGTM. Thanks Rob!

PR #693, opened 2021-01-04
Backport: Update Guice config to split forwarder and server (#686)

danotorrey:
Backport #686 into `cloud-4.0` so that it can be used with the current Cloud code.

I suppose there is no need to also backport into `4.0`, since this is only needed for the Forwarder, which. is only used with `cloud-4.0`


waab76 (approved):
LGTM

PR #719, opened 2021-02-03
Initial version of USERID parsing for Palo9x input

waab76:
This change adds support for the `USERID` log type in the Palo Alto 9.x input.  This is needed for upcoming work on Graylog Illuminate.

JUnit tests have been added to ensure the new message type is handled as expected.

Note: there is currently an open question regarding the `User Group Flags` (position 32) and `User by Source` (position 32) fields in the `USERID` log type.  Per the Palo schema, positions 30 and 31 should be reserved for future use.  However, test logs obtained both from SumoLogic and from Nick's own Palo firewall seem to show `User Group Flags` in position 30 and `User by Source` in position 31.  We need to resolve this discrepancy before releasing this update.


cobaltclaudia (approved):
LGTM

Issue #726, opened 2021-02-08
Palo 9.x Input performance deep dive


waab76:
A user has reported that when they turn on the Palo Alto 9.x input, their system begins to fall behind on processing.  When we looked at the user's node metrics, we saw that the p95 `parseTime` metric for the Palo input was 860 usecs and they were receiving roughly 2000 messages per second.  This works out to roughly 1.7 seconds worth of work coming in every second and thus the processing backlog is not surprising.

We want to investigate where the Palo 9.x input is spending the time covered in the `parseTime` metric and if there is a way we could be processing more efficiently or if we have simply hit a performance limit on the system.

waab76:
Ran a bunch of messages through a Palo 9.x input locally and got the following numbers:

<table>
<tr>
	<td>Metric
	<td>Mean
	<td>p95
	<td>p99
<tr>
	<td>Total Parse Time
	<td>2056 μs
	<td>2595 μs
	<td>9690 μs
<tr>
	<td>CSV Split Time
	<td>62 μs
	<td>159 μs
	<td>485 μs
<tr>
	<td>Template Handling Time
	<td>799 μs
	<td>2217 μs
	<td>4580 μs
</table>

There were 1.6MM events processed at an average of 820 events per second.

<details>
<summary>Metrics Screenshot</summary>

![Screen Shot 2021-02-08 at 12 42 45 PM](https://user-images.githubusercontent.com/6466251/107260294-f99e0780-6a0b-11eb-89e3-e8c8bf1b49d9.png)

</details>

My additional instrumentation obviously isn't capturing everything related to the total `parseTime` metric, but it does seem that the bulk of the time is going into template handling.  I'm going to take a closer look at that code to see if there are opportunities for optimization.

waab76:
Made some changes to templates and parser to parse fields in order rather than semi-random. Seems to have made a positive change:



<table>
<tr>
	<td>Metric
	<td>Mean
	<td>p95
	<td>p99
<tr>
	<td>Total Parse Time
	<td>158 μs
	<td>247 μs
	<td>1063 μs
<tr>
	<td>CSV Split Time
	<td>78 μs
	<td>111 μs
	<td>776 μs
<tr>
	<td>Template Handling Time
	<td>70 μs
	<td>108 μs
	<td>364 μs
</table>

There were 400K events processed at an average of 840 events per second.

<details>
<summary>Metrics Screenshot</summary>

[...]

PR #731, opened 2021-02-11
Refactor PaloAltoTypeParser to improve performance

waab76:
As documented in #726, a customer reported perfomrance issues with the Palo 9.x input.  

This change is an attempt to improve Palo input parse times by ensuring the Palo template fields are ordered by position and iterating over the input string from beginning to end rather than going through the unordered template fields and essentially doing random access on the input string.

Verified that the changes do not break existing Palo JUnits.  Also tested locally with the data driver script and sample Palo log files.

### DIFF
--- a/src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
+++ b/src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
@@ -35,11 +35,8 @@ public class PaloAltoTypeParser {
     private static final Logger LOG = LoggerFactory.getLogger(PaloAltoTypeParser.class);
 
     private final PaloAltoMessageTemplate messageTemplate;
-    private final PaloAltoMessageType messageType;
 
-    public PaloAltoTypeParser(PaloAltoMessageTemplate messageTemplate, PaloAltoMessageType messageType) {
-
-        this.messageType = messageType;
+    public PaloAltoTypeParser(PaloAltoMessageTemplate messageTemplate) {
         this.messageTemplate = messageTemplate;
     }
 
@@ -80,7 +77,8 @@ public class PaloAltoTypeParser {
 
                 // Handling of duplicate keys
                 if (fieldMap.containsKey(template.field())) {
-                    if (Strings.isNullOrEmpty(rawValue.trim()) || value.equals(fieldMap.get(template.field()))) {
+                    if (Strings.isNullOrEmpty(rawValue.trim()) || null == value
+                            || value.equals(fieldMap.get(template.field()))) {
                         // Same value, do nothing
                     } else if (fieldMap.get(template.field()) instanceof List) {
                         List valueList = (List) fieldMap.get(template.field());
--- a/src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
+++ b/src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
@@ -26,6 +26,10 @@ import org.slf4j.LoggerFactory;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
```diff
@@ -41,68 +45,62 @@ public PaloAltoTypeParser(PaloAltoMessageTemplate messageTemplate, PaloAltoMessa
...
         Map<String, Object> fieldMap = Maps.newHashMap();
+        List<PaloAltoFieldTemplate> templateFields = Lists.newArrayList(messageTemplate.getFields());
 
-        for (PaloAltoFieldTemplate field : messageTemplate.getFields()) {
-            String rawValue = null;
-            try {
-                rawValue = fields.get(field.position());
-            } catch (IndexOutOfBoundsException e) {
-                // Skip fields at indexes that do not exist.
-                LOG.trace("A [{}] field does not exist at index [{}]", messageType.toString(), field.position());
```
bernd, line 51: The `messageType` field is now unused.

src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
```diff
@@ -41,68 +45,62 @@ public PaloAltoTypeParser(PaloAltoMessageTemplate messageTemplate, PaloAltoMessa
...
+                } else if (template.fieldType() == BOOLEAN) {
+                    value = Boolean.valueOf(rawValue);
+                } else {
+                    LOG.warn("Unrecognized data type [{}] for field [{}], handling as STRING",
+                            template.fieldType(), template.field());
+                }
+
+                // Handling of duplicate keys
+                if (fieldMap.containsKey(template.field())) {
+                    if (Strings.isNullOrEmpty(rawValue.trim()) || value.equals(fieldMap.get(template.field()))) {
```
bernd, line 83: The IDE warns about a possible NPE with `value.equals(...)`. Not sure if that can happen in practice.

bernd (changes requested):
Two small nits:

src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
waab76: Good catch

src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoTypeParser.java
waab76: `value` could be null if `rawValue` was null, in which case the `Strings.isNullOrEmpty(..)` check would catch it.  It could also be null if we caught a NumberFormatException trying to parse a Long.  I'll add a null check just to be safe.

bernd (approved):
LGTM :+1: 

One question: We are still parsing the templates for every message, even if the default templates will be used.

https://github.com/Graylog2/graylog-plugin-integrations/blob/68bb7ddbfd54af097bdeec22a60fc68015f9fd52/src/main/java/org/graylog/integrations/inputs/paloalto/PaloAltoCodec.java#L55-L57

Would that be another possible optimization?

waab76:
@bernd I haven't looked too closely at performance enhancements to the Palo 8.x input (which is what uses the `PaloAltoCodec`) since it was the 9.x input that the customer was reporting problems with.  Since users have the ability to make changes to the templates for 8.x, we need to ensure that we don't miss a change.  It might make sense to move the template construction into the Input object so that we don't miss a config change and then we'd need to inject the constructed templates into the Codec.

I'll take a look at it.

bernd:
@waab76 Ah right! Sorry, forgot about that again... :smile: I think we should merge this. If we find other improvements we can just do them in other PRs.

PR #736, opened 2021-02-16
Support AWS custom Input UI in Forwarder

edmundoa:
This PR adapts the existing AWS custom Input UI for the Forwarder. The main changes consist in:

- Making it possible to render the UI embedded into another page, instead of using a separate page for it
- Bypass the Input submission through the plugin API, and instead let the Forwarder create the Input itself. The regular Input creation will still work as it did before
- Add frontend tests and configure them in the project
- Fix some smaller linter issues

Needs https://github.com/Graylog2/graylog-plugin-cloud/pull/741 for the forwarder input UI to work.

/jenkins-pr-deps Graylog2/graylog2-server#10072,Graylog2/graylog-plugin-cloud#741

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,19 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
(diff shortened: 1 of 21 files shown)


edmundoa:
@cobaltclaudia @waab76 I added you for review since you have more knowledge about this code than we do. Let me know if you see any concerns with the general approach as well.

We could send the final form data to the AWS plugin backend if we needed to do some extra verification or transformation of the data (all the Forwarder code cares about is receiving an Input back that can be saved), but I thought this was not needed at this point.

I think this Input should run on the cloud side at some point, since it seems that running it on the Forwarder is an unnecessary redirect. At the moment we have no good way of achieving that, so this is still better than nothing.

ousmaneo (approved):

waab76 (approved):
I have to admit, I'm not super familiar with this code.  That said, I don't see any obvious issues.

ousmaneo:
Merged. thank @edmundoa

PR #760, opened 2021-03-05
Greynoise

cobaltclaudia:
Greynoise IP Context Data adapter

Config
Screen Shot 2021-03-29 at 3 40 21 PM

Lookup Result
Screen Shot 2021-03-09 at 5 03 25 PM

### DIFF
--- a/src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
+++ b/src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
@@ -24,6 +24,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Maps;
 import com.google.inject.assistedinject.Assisted;
 import com.unboundid.util.json.JSONException;
 import com.unboundid.util.json.JSONObject;
@@ -42,10 +44,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
 import javax.validation.constraints.NotEmpty;
-import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
@@ -56,24 +55,23 @@ public class GreyNoiseDataAdapter extends LookupDataAdapter {
 
     private final EncryptedValueService encryptedValueService;
     private final Config config;
-    private OkHttpClient okHttpClient;
+    private final OkHttpClient okHttpClient;
 
     @Inject
     public GreyNoiseDataAdapter(@Assisted("id") String id,
                                 @Assisted("name") String name,
                                 @Assisted LookupDataAdapterConfiguration config,
                                 MetricRegistry metricRegistry,
-                                EncryptedValueService encryptedValueService) {
+                                EncryptedValueService encryptedValueService,
+                                OkHttpClient okHttpClient) {
         super(id, name, config, metricRegistry);
         this.config = (Config) config;
         this.encryptedValueService = encryptedValueService;
+        this.okHttpClient = okHttpClient;
     }
 
     @Override
     public void doStart() throws Exception {
-
-        okHttpClient = new OkHttpClient().newBuilder()
-                                         .build();
 
     }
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+                    .addHeader("Accept", "application/json")
+                    .addHeader("key", encryptedValueService.decrypt(config.apiToken()))
+                    .build();
+
+            return parseResponse(okHttpClient.newCall(request).execute());
+
+        } catch (Exception e) {
+            LOG.error("An error occurred while retrieving lookup result [{}]", e.toString());
+            return LookupResult.withError();
+        }
```
waab76, line 111: It's generally a good practice to ensure the `Response` object returned from `.execute()` gets closed.  This is a good place to use Java's try-with-resources functionality:

```suggestion
        Request request = new Request.Builder()
                .url(GREYNOISE_IPQC_ENDPOINT + keyObject.toString())
                .method("GET", null)
                .addHeader("Accept", "application/json")
                .addHeader("key", encryptedValueService.decrypt(config.apiToken()))
                .build();

        try (Response response = okHttpClient.newCall(request).execute()) {

            return parseResponse(response);

        } catch (Exception e) {
            LOG.error("An error occurred while retrieving lookup result [{}]", e.toString());
            return LookupResult.withError();
        }
```

Declaring the `Response` object in `try (...) {` will ensure it automatically gets closed when the try/catch block finishes.

src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    public static LookupResult parseResponse(Response response) throws IOException {
+
+        if (response.isSuccessful()) {
+            Map<Object, Object> map = new HashMap<>();
```
waab76, line 117: I tend to prefer Guava's `Maps.newHashMap()` just because I consider use of the `new` keyword to be a code smell.

src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+
+            BufferedReader in = new BufferedReader(new InputStreamReader(Objects.requireNonNull(response.body()).byteStream()));
+            String jsonString = in.readLine();
```
waab76, line 120: Is there a reason you didn't use `response.body().string()`?

src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+
+            return parseResponse(okHttpClient.newCall(request).execute());
+
+        } catch (Exception e) {
+            LOG.error("An error occurred while retrieving lookup result [{}]", e.toString());
+            return LookupResult.withError();
+        }
+    }
+
+    public static LookupResult parseResponse(Response response) throws IOException {
```
waab76, line 114: This doesn't really need to be public.  You can remove the `public` modifier to leave it package private and add a `@VisibleForTesting` annotation to explain why it's not `private`

Also, does it need to be static or is that just to make the test easier to write?

src/test/java/org/graylog/integrations/dataadapters/GreyNoiseDataAdapterTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.assertj.core.api.Assertions;
+import org.graylog.integrations.dataadpaters.GreyNoiseDataAdapter;
+import org.graylog2.plugin.lookup.LookupResult;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class GreyNoiseDataAdapterTest {
```
waab76, line 33: I think once you move the `OkHttpClient` into the constructor, you can mock it out and that will allow you to write a couple more test cases such as:

* Verifying that the expected URL is being used to make the call
* Verifying the behavior when the HTTP request isn't successful

src/web/dataadapters/GreyNoiseAdapterDocumentation.jsx
```diff
@@ -0,0 +1,27 @@
...
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+/* eslint-disable react/no-unescaped-entities, no-template-curly-in-string */
+import React from 'react';
+
+class GreyNoiseAdapterDocumentation extends React.Component {
+  render() {
+    return () => {
+    };
```
waab76, line 23: We should have something here.  Maybe a quick adaptation of some text from GreyNoise's own documentation.

waab76 (changes requested):

src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
cobaltclaudia: to make it easier to write the test

src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
cobaltclaudia: no, was an oversight

src/main/java/org/graylog/integrations/dataadpaters/GreyNoiseDataAdapter.java
waab76: Ok, I don't think there's any danger to keeping it `static` since it doesn't interact with anything outside of its own scope.  But I do think it's worth dropping the `public` modifier. The JUnit shouldn't have any problem accessing the method since they're in the same package.

```suggestion
    static LookupResult parseResponse(Response response) throws IOException {
```

waab76 (changes requested):
Last nit regarding the `public` modifier on the parse method

waab76 (approved):
LGTM.  Thanks for making all the requested changes.

PR #798, opened 2021-04-12
Hide global input option for AWS on cloud

Gaya:
As described here: https://github.com/Graylog2/graylog-plugin-cloud/issues/763

We need to hide the global input option on cloud instances.


thll:
Sorry that this hasn't been merged but apparently it got fixed elsewhere.

Issue #816, opened 2021-04-28
Invalid format using PAN OS input


edward-marrufo:
## Expected Behavior

When attempting to create an input in Graylog using PAN OS 9 input I would expect to see it work with the provided dates in the message.

## Current Behavior

Currently the first time stamp is working normally and the second time stamp (event created) shows "invalid format" in the server logs.
## Steps to Reproduce (for bugs)

1. See HS-382974779 for supporting information.

bernd:
I moved this issue from the server repository.

kmerz:
I did some research on this issue since I suspected a wrong pipeline rule.

After some research I found that the PaloAlto9x input is overriding the timestamp field here:

https://github.com/Graylog2/graylog-plugin-integrations/blob/master/src/main/java/org/graylog/integrations/inputs/paloalto9/PaloAlto9xCodec.java#L85

which leads to the problem that the `timestamp` field is no longer a DateTime object but the plain string,
as it can be seen in the test:

https://github.com/Graylog2/graylog-plugin-integrations/blob/master/src/test/java/org/graylog/integrations/inputs/paloalto9/PaloAlto9xTemplatesTest.java#L86

The user sees this problem when he tries to parse the timestamp field in a pipeline rule:
```
2021-04-21T16:03:23.034+02:00 WARN  [ProcessBufferProcessor] Unable to process message <55aaf7a0-a2aa-11eb-9624-005056a1a599>: org.graylog.plugins.pipelineprocessor.ast.exceptions.FunctionEvaluationException: java.lang.IllegalArgumentException: Invalid format: "2021/04/21 16:03:08" is malformed at "/04/21 16:03:08" at
```

Since he could not send any message to the system it is unclear to me what happens after the pipeline processing. Perhaps there some cleanup with the timestamp is made.

waab76:
The error is getting thrown from this line of code: https://github.com/Graylog2/graylog2-server/blob/1b84172433551fcf53fb91cb190685f74[…]raylog/plugins/pipelineprocessor/functions/dates/ParseDate.java

Maybe it wouldn't be going down to that line of code if the timestamp field were stored as a `DateTime` instead of a `String`, but the fact remains that the parsing (of what looks like a valid timestamp string) is failing

I had a suspicion that it was failing due to the timestamp string not matching the expected pattern for the locale, but I have no idea where the locale used in `ParseDate` is coming from

But in all our testing (presumably using en_US as the locale), this hasn't been an issue

[...]

PR #825, opened 2021-05-07
Ensure PAN timestamps are DateTime not String (#816)

waab76:
Per issue #816, customers have noticed issues with the `timestamp` field in PAN 9.x inputs.  This change ensures that the `timestamp` field gets converted from `String` to `DateTime` in the codec.

This has been tested locally with JUnit tests.


kmerz (approved):
Fix works for me :heavy_check_mark:

Issue #835, opened 2021-05-13
Add MatterMost to URL regex for Slack plugin


kroepke:
# Description
MatterMost is also Slack webhook compatible so we should allow using it. 

## What
The config for the slack notification aggressively checks for domain names as part of its URL validation, so you cannot enter anything else than slack and discord. 
Either just add MatterMost or simply allow any host name in the URL. 

## Why
User request at https://twitter.com/komodin/status/1392958577346809862?s=21

kroepke:
As per the Twitter conversation, consider using the URL class and check for the expected components only using the path components. The host should probably not matter so we don't have to check for specifics (unless there are differences in behavior, of course).

PR #836, opened 2021-05-14
Only validate Slack Notif webhook URL if it is a Slack or Discord URL

waab76:
Per #835, the Slack Notification plugin is currently only allowing Slack and Discord webhook URLs when there are other messaging systems capable of accepting Slack-formatted input.  This change updates the Slack Notification plugin to do basic URL validation on provided webhook URLs unless those URLs include "slack" or "discord" in the domain name.  Webhook URLs with "slack" or "discord" in the domain name, will be further validated against the existing Slack/Discord patterns.

JUnit test has been updated to cover all new cases.


cobaltclaudia (approved):
LGTM

PR #925, opened 2021-08-18
Parse timestamps in PaloAltoTypeParser

mpfz0r:
This moves the timestamp parsing to an earlier stage in the processing,
and avoids setting an intermediate String to a Message timestamp field.

The changes introduced in
 https://github.com/Graylog2/graylog2-server/pull/11149
are replacing unparseable timestamps with the current time.
Thus the previous approach of fixTimestampField() will not
find the earlier set String timestamp anymore.

Refs https://github.com/Graylog2/graylog-plugin-integrations/pull/825

## Notes for Reviewers


waab76 (approved):
LGTM

PR #927, opened 2021-09-09
Make slack channel name templatable

jjpp-sympower:
We have support teams in different countries that follow dedicated slack channels. Slack channel names can be automatically derived from the same data as the custom message can. As of now we have to create an alerting rule per country where we could have just one rule. This PR implements templating the channel name using the same model as the custom message.


alaraun:
Looks good and makes life much easier.
@waab76 maybe you can review :)

PR #939, opened 2021-11-01
Renames OS GreyNoise adapter and adds User-Agent header

kingzacko1:
- Renames GreyNoise Data Adapter to GreyNoise Quick IP Lookup 
- Adds generic "Graylog/Version" User-Agent header to GreyNoise API call - open to suggestions on what else to add to that header

## Notes for Reviewers

### DIFF
--- a/src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
+++ b/src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
@@ -32,6 +32,7 @@ import com.unboundid.util.json.JSONObject;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.graylog2.plugin.Version;
 import org.graylog2.plugin.lookup.LookupCachePurge;
 import org.graylog2.plugin.lookup.LookupDataAdapter;
 import org.graylog2.plugin.lookup.LookupDataAdapterConfiguration;
@@ -48,8 +49,8 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.Objects;
 
-public class GreyNoiseDataAdapter extends LookupDataAdapter {
-    private static final Logger LOG = LoggerFactory.getLogger(GreyNoiseDataAdapter.class);
+public class GreyNoiseQuickIPDataAdapter extends LookupDataAdapter {
+    private static final Logger LOG = LoggerFactory.getLogger(GreyNoiseQuickIPDataAdapter.class);
     public static final String NAME = "GreyNoise";
     static final String GREYNOISE_IPQC_ENDPOINT = "https://api.greynoise.io/v2/noise/quick/";
 
@@ -58,12 +59,12 @@ public class GreyNoiseDataAdapter extends LookupDataAdapter {
     private final OkHttpClient okHttpClient;
 
     @Inject
-    public GreyNoiseDataAdapter(@Assisted("id") String id,
-                                @Assisted("name") String name,
-                                @Assisted LookupDataAdapterConfiguration config,
-                                MetricRegistry metricRegistry,
-                                EncryptedValueService encryptedValueService,
-                                OkHttpClient okHttpClient) {
+    public GreyNoiseQuickIPDataAdapter(@Assisted("id") String id,
+                                       @Assisted("name") String name,
+                                       @Assisted LookupDataAdapterConfiguration config,
+                                       MetricRegistry metricRegistry,
+                                       EncryptedValueService encryptedValueService,
+                                       OkHttpClient okHttpClient) {
         super(id, name, config, metricRegistry);
         this.config = (Config) config;
         this.encryptedValueService = encryptedValueService;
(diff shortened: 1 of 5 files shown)


danotorrey:
Closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/2885

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
```diff
@@ -120,22 +125,22 @@ static LookupResult parseResponse(Response response) {
             } catch (JSONException | IOException e) {
                 LOG.error("An error occurred while parsing Lookup result [{}]", e.toString());
             }
-            return LookupResult.withoutTTL().multiValue(map)
-                               .build();
-        } else
+            return LookupResult.withoutTTL().multiValue(map).build();
+        } else {
             return LookupResult.empty();
+        }
```
kingzacko1, line 131: This must've been updated with the auto-format on save according to the Graylog formatting profile

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
```diff
@@ -48,22 +49,22 @@
 import java.util.Map;
 import java.util.Objects;
 
-public class GreyNoiseDataAdapter extends LookupDataAdapter {
-    private static final Logger LOG = LoggerFactory.getLogger(GreyNoiseDataAdapter.class);
-    public static final String NAME = "GreyNoise";
```
danotorrey, line 53: To maintain backwards compatibility, the adapter name constant cannot be renamed, since it is saved in the Mogno `db.lut_data_adapters` collection. You can find this with the Mongo CLI in the `config.type` field:
```
$ mongo
> use graylog
> db.lut_data_adapters.find()
``` 

```
{
        "_id" : ObjectId("617ff5032fcd864f475fc7f3"),
        "title" : "Test",
        "description" : "test",
        "name" : "test",
        "custom_error_ttl_enabled" : false,
        "config" : {
                "type" : "GreyNoise",
                "api_token" : {
                        "encrypted_value" : "53454353253454325",
                        "salt" : "342342"
                }
        }
}
```

Changing it would result in an error for adapters that exist already, so I think it needs to stay the same.
image

The class name changes should be OK, I think. Do you agree @waab76 @bernd?

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
```diff
@@ -92,11 +93,15 @@ protected void doRefresh(LookupCachePurge cachePurge) throws Exception {
 
     @Override
     protected LookupResult doGet(Object keyObject) {
+        String userAgent = String.format("Graylog/%d.%d",
+                Version.CURRENT_CLASSPATH.getVersion().getMajorVersion(),
+                Version.CURRENT_CLASSPATH.getVersion().getMinorVersion());
```
danotorrey, line 98: I think just using the string representation of the version object would return the full version, which would be a bit simpler. Thinking of the server startup message example:

https://github.com/Graylog2/graylog2-server/blob/10c3095e40e30f65f2c5936a968a98b05fc69e7a/graylog2-server/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java#L128

> Graylog server 4.3.0-SNAPSHOT starting up

danotorrey (changes requested):

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
kingzacko1: Reverted back to previous NAME for backwards compatibility

danotorrey: Thanks. I agree. I think renaming them makes sense.

danotorrey:
I think it would also help to add the in progress issues/PRs to the [project board](https://github.com/orgs/Graylog2/projects/45) in the `In Work` column. I am trying to remember to do this for my PRs too.

kingzacko1:
I had the issue "In Work" but once I opened the PR for it I moved it over to "In Review". I tried to link the PR to our Project but I could only see the one we tagged with "--- OLD" and not the more recent one. Not sure why that is...

danotorrey:
@kingzacko1 Ahh, I think that's because the Integrations project is currently private, and this is an open-source repo. I will ask in the team if we can make it public.

Another question: I am trying to view the most recent changes from the PR feedback, but I am seeing this force-commit, which shows a bigger change-set. I am just wondering if you know why the bigger change set is showing. Is the force-push needed for a particular reason?

image

danotorrey:
> @kingzacko1 Ahh, I think that's because the Integrations project is currently private, and this is an open-source repo. I will ask in the team if we can make it public.

Hmm, it's not because it's private... there are other private projects in the organization list that are viewable. I'll keep investigating...

danotorrey:
@kingzacko1 Are you able to see the Integrations project on the Organization tab for this PR? I see it on my recent tab, but not the organization tab.

danotorrey (approved):
LGTM and tests successfully.

cobaltclaudia (approved):
LGTM

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseQuickIPDataAdapter.java
kingzacko1: > @kingzacko1 Are you able to see the Integrations project on the Organization tab for this PR? I see it on my recent tab, but not the organization tab.

Checked it out this morning and it was there and available on the Organizations tab... but it was nowhere to be found yesterday - recent nor organization

PR #953, opened 2021-12-17
Updated GreyNoiseCommunityIpLookupAdapter to exclude 'message' field,…

roberto-graylog:
Updated GreyNoise Community IP Lookup Adapter.  Included here is my proposal for handling/checking reserved IP addresses.  Let me know what you think--I am all ears.

## Notes for Reviewers


src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java
```diff
@@ -125,29 +131,45 @@ protected LookupResult doGet(Object ipAddress) {
...
 
+    private String getValidIpString(Object ipAddress) {
+        String ipString = ipAddress == null ? "" : ipAddress.toString();
+        if (ipString.trim().isEmpty()) {
+            String error = String.format("[%s] requires an IP address to perform Lookup", ADAPTER_NAME);
+            throw new IllegalArgumentException(error);
+        }
+
+        if (!InetAddressValidator.getInstance().isValidInet4Address(ipString)) {
+            String error = String.format("[%s] is not a valid IPv4 Address", ipString);
```
kingzacko1, line 161: Should we also do a check to see if the IP is IPv6 and not throw an error if it is IPv6, but maybe just print out a debug log saying that no GreyNoise lookup will be attempted since the IP is IPv6? Throwing an error for an IPv6 address seems wrong since it technically isn't an invalid IP, it's just not a supported IP.

kingzacko1 (changes requested):

kingzacko1:
Just realized something else while trying to use the new ReservedIPChecker class into the enterprise adapter. The graylog-plugin-enterprise-integrations project does not currently have a dependency on graylog-plugin-integrations. Would it make sense to move the ReservedIpChecker class to graylog2-server project possibly in the utilities package so 1) we don't have to introduce a new dependency and 2) the code can be easily reused anywhere else in our codebase?

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java
roberto-graylog: Yes, I was wondering the same for a reserved IP address.  As is, the result (in the test screen) is very unhelpful:
```
{
  "single_value": null,
  "multi_value": null,
  "string_list_value": null,
  "has_error": true,
  "ttl": 5000
}```

I am not sure that we need to check whether it is IPv6, since we really only care that it is a valid IPv4--unless we want to be extra helpful, which may not be a bad idea?

**Comment by roberto-graylog** [comment]
>

File: src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java (hunk 1 above)
Comment (kingzacko1) [reply src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java:161]: I recommend checking for IPv6 so that we can safely ignore them and not throw an error since it is perfectly valid to have an IPv6 address, we just wont send it to GreyNoise. On the off chance a user overrwrites an IP address field with an invalid value or has them sent in/parsed incorrectly I'd think we'd want different behavior than if an IPv6 address came in. And I agree, it should work the same for non-routable IPs.

Comment (roberto-graylog) [reply src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java]: Thanks for the feeback @kingzacko1 .  I have addressed your points.  I created a PR ([#11873](https://github.com/Graylog2/graylog2-server/pull/11873)) for the reserved IP checker.

File: src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java [hunk 2]
```diff
@@ -125,27 +136,53 @@ protected LookupResult doGet(Object ipAddress) {
...
+        final String ipString = ipAddress == null ? "" : ipAddress.toString();
+        if (ipString.trim().isEmpty()) {
+            String error = String.format("'%s' requires an IP address to perform Lookup", ADAPTER_NAME);
+            throw new IllegalArgumentException(error);
+        }
+
+        final String validIpAddress;
+        final InetAddressValidator validator = InetAddressValidator.getInstance();
+        if (validator.isValidInet6Address(ipString)) {
+            LOG.error("'{}' is an IPv6 Address.  '{}' does not support IPv6 Addresses", ipAddress, ADAPTER_NAME);
```
kingzacko1, line 173: Do you think Reserved and IPv6 addresses may be better off being logged at a lesser severity - `WARN` or possibly even `INFO`? There is nothing wrong with those IPs coming through Graylog we just aren't going to check them with GreyNoise because there is no need. Spitting out an error log message for that seems excessively severe. 

Invalid IPv4 can still be logged at the error level since that would definitely be a problem. We may also want to look at possibly including an error message in the lookup if that does happen. We should be able to merge it without it, but it may be beneficial to give some information about why there was a lookup error in the result itself if that is possible... I haven't worked with look up tables enough to know if we handle something like that already.

kingzacko1 (changes requested):

src/main/java/org/graylog/integrations/dataadapters/GreyNoiseCommunityIpLookupAdapter.java
roberto-graylog: I see what you mean.  My point of view was that, from the adapter itself, IPv6 is an invalid input-error.  @danotorrey , do you have any thoughts? Is this better off as a **Warning**?

I discussed the message with @waab76 .  What (I understood) he said is that the message is not really useful in a lookup table as the user has no good way of seeing it--the user only cares whether there is a value.

kingzacko1 (approved):
LGTM

PR #965, opened 2022-03-22
Don't follow redirects when calling Slack API

thll:
The Slack API is issuing 302 redirects for unknown
webhook URLs. The eventual target will return a 200,
so the http client will treat them as successful

By turning off redirects for slack notifications, we
can detect these cases as failures.

Fixes https://github.com/Graylog2/graylog2-server/issues/12315


patrickmann (approved):
Looks good and works as expected

PR #987, opened 2022-04-04
OpsGenie_Alert_Notification

Srinidhi-Saravanan:
- ## Notes for Reviewers
- Created new plugin to send Graylog Alert message to OpsGenie.
- Created user notification config page.

- [x] The commit history must be preserved - please use the rebase-merge or standard merge option instead of squash-merge
- [x] Sync up with the author before merging


danotorrey:
We should be able to port this code over and test it out with a trial OpsGenie account. As far as I can tell, the `https://api.opsgenie.com/v2/alerts` endpoint should work within the 14 day trial.

rich-graylog:
While revisiting this PR, we noticed that OpsGenie is actually approaching EOL, so we may want to look at Atlassian Jira Svc Mgmt instead.  More info here: https://www.atlassian.com/blog/announcements/evolution-of-it-operations

PR #1069, opened 2022-06-15
Refactor slack and teams notifications

kingzacko1:
refactor of the Slack and Teams notification types code to more closely align with use of AutoValue across the project. There should be no functional difference in the two notifications.

## Notes for Reviewers


ryan-carroll-graylog (changes requested):
The original functionality is working in testing.

I did notice an existing bug where you can use a teams webhook for a slack notification and vice versa. I think combining the nested if statements in the following files should fix the issue:
https://github.com/Graylog2/graylog-plugin-integrations/blob/07ca26e6a4cf87a5c5d14cdbf8cf5a662214a1e7/src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java#L125-L130

https://github.com/Graylog2/graylog-plugin-integrations/blob/07ca26e6a4cf87a5c5d14cdbf8cf5a662214a1e7/src/main/java/org/graylog/integrations/notifications/types/microsoftteams/TeamsEventNotificationConfig.java#L97-L98

src/main/java/org/graylog/integrations/notifications/types/SlackClient.java
```diff
@@ -52,16 +56,17 @@ public SlackClient(OkHttpClient httpClient) {
...
                 .url(webhookUrl)
-                .post(RequestBody.create(MediaType.parse(APPLICATION_JSON), message.getJsonString()))
+                .post(RequestBody.create(MediaType.parse(APPLICATION_JSON), objectMapper.writeValueAsString(message)))
                 .build();
 
-        LOG.debug("Posting to webhook url <{}> the paylod is <{}>",
+        LOG.debug("Posting to webhook url <{}> the payload is <{}>",
                 webhookUrl,
-                message.getJsonString());
+                "");
```
ryan-carroll-graylog, line 69: Was this supposed to be `objectMapper.writeValueAsString(message` ?

kingzacko1: yup, good catch. I just blew away the original parameter to get things to compile originally

kingzacko1:
Fixed the log message and updated URL validation logic in latest commit @ryan-carroll-graylog

src/main/java/org/graylog/integrations/notifications/types/microsoftteams/TeamsEventNotificationConfig.java
```diff
@@ -91,16 +92,13 @@ public static Builder builder() {
...
-                    validation.addError(FIELD_WEBHOOK_URL, INVALID_TEAMS_URL_ERROR_MESSAGE);
-                }
+            final URI webhookUri = new URI(webhookUrl());
+            if (!webhookUri.getHost().toLowerCase().contains("office") ||
+                    !TEAMS_PATTERN.matcher(webhookUrl()).find()) {
+                validation.addError(FIELD_WEBHOOK_URL, INVALID_TEAMS_URL_ERROR_MESSAGE);
             }
-
-        } catch (Exception ex) {
+        } catch (URISyntaxException ex) {
```
ryan-carroll-graylog, line 101: This now handles an incorrect webhook perfectly, but I noticed here that if you use something crazy that's not a valid URI, you get a `NullPointerException` at line 97 which isn't caught anymore. The one downside is that you end up with an 'Unknown error' at the UI instead of explicitly getting a pretty 'Invalid URL' message with field highlighting.

Same with the slack notification.

kingzacko1: Yep, fixed that and reverted back to a generic Exception

ryan-carroll-graylog: > Yep, fixed that and reverted back to a generic Exception

You are a beast, I can't even think as fast as you crank out commits

ryan-carroll-graylog (approved):

ryan-carroll-graylog (approved):

roberto-graylog (approved):
Looks good.

danotorrey:
Thank you @kingzacko1!

PR #1077, opened 2022-06-22
Migrate PagerDuty Notifications from v1 to v2

### LINKED ISSUE #1054: Missing migration from v1 config in pagerduty notification
The Pagerduty notification code was migrated in October 2020 from https://github.com/graylog-labs/graylog-plugin-pagerduty, and its config version bumped from `v1` to `v2` in the process.

There doesn't seem to be a migration for `v1` configs, so if a very old config is still present, this can lead to all kinds of issues whenever the entity is loaded.
In this case, a share dialog tried to load the `v1` config from the database and failed.

I think we need to add a v1 to v2 migration as a bug fix if that is possible. If it's not possible, we should fail more gracefully.

See https://github.com/Graylog2/graylog-plugin-integrations/blob/08a16fed7039c5e4c1bbdc269fa36dc458d91f1f/src/main/java/org/graylog/integrations/pagerduty/PagerDutyNotificationConfig.java#L47

kingzacko1:
## Notes for Reviewers
Updates the `type.config` field for PagerDuty notifications.
closes #1054 

### DIFF
--- a/src/main/java/org/graylog/integrations/IntegrationsModule.java
+++ b/src/main/java/org/graylog/integrations/IntegrationsModule.java
@@ -35,6 +35,7 @@ import org.graylog.integrations.inputs.paloalto9.PaloAlto9xInput;
 import org.graylog.integrations.ipfix.codecs.IpfixCodec;
 import org.graylog.integrations.ipfix.inputs.IpfixUdpInput;
 import org.graylog.integrations.ipfix.transports.IpfixUdpTransport;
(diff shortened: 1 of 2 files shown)


roberto-graylog (approved):

ryan-carroll-graylog (approved):
Looks good. Tested with V1 config and it was successfully migrated to V2, enabling loading without failure.

bernd:
@kingzacko1 Can this be merged? :slightly_smiling_face:

kingzacko1:
@bernd for some reason I thought it already had been, all done!

PR #1115, opened 2022-07-22
Minor changes for Jackson 12.3.3

patrickmann:
We are updating Jackson to a current version, so we no longer need the dependency exclusion.
Also includes an update of the Jest config, required by the [updated Jest client](https://github.com/graylog-labs/Jest/pull/4) for ES6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "build": "webpack",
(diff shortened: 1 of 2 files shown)


patrickmann:
Postponed to 5.0, when we expect to drop ES6.

PR #1163, opened 2022-09-06
Migrate PagerDuty Notifications from v1 to v2 (#1077)

kingzacko1:
Backport of #1077 
## Notes for Reviewers

- [ ] The commit history must be preserved - please use the rebase-merge or standard merge option instead of squash-merge
- [ ] Sync up with the author before merging

PR #1167, opened 2022-09-08
Don't add migrations to the forwarder bindings

mpfz0r:
As you can see on the comment on top of configureUniversalBindings(), the forwarder does not have MongoDB.

Refs #1163


kingzacko1 (approved):
Didn't notice the separate configure methods when doing this one, sorry about that.

PR #1202, opened 2022-10-14
Rework Microsoft Teams Notification Templating

### LINKED ISSUE #1096: Microsoft Teams notification type `getMessageDetails` incorrectly builds content (missing data)
## Description

The Microsoft Teams notification type's `getMessageDetails()` incorrectly handles splitting on `:`.  The most obvious result of this is fact "values" that contain `:` (e.g. all timestamps and messages in the backlog section) only have the value left of that `:` (e.g. `2022-06-30T19:12:34.567Z` becomes `2022-06-30T19`).

A bare minimum fix for the basics would be `field.split(":", 2);`, but even in the default custom message there is content where that is inappropriate (e.g. `Event Fields` and `Backlog`).

image

## Steps To Reproduce

1. Create a `Microsoft Teams Notification` notification type with a valid Teams webhook URL and the default template
1. Use that notification on an event definition that includes backlog messages (and can be easily triggered)
1. Note that the notification posted to the Teams channel is missing timestamp and backlog content

## Environment
- Graylog Version:  4.3.2

### LINKED ISSUE #1200: MS Teams Notification is broken (timestamp)
This is related to this thread:
https://community.graylog.org/t/ms-teams-notification-broken-for-me/24563

## Expected Behavior
Show timestamp: "2022-10-12 12:16:50.252"

## Current Behavior
It shows this: "2022-10-12 12"

## Possible Solution
Fix the colon scape.

## Steps to Reproduce (for bugs)
Create a new webhook in teams and set an alert for any event that has a timestamp like "2022-10-12 12:16:50.252".

## Context
It's very annoying to not be able to see the correct minute, second and millisecond that something happened.

## Your Environment
* Graylog Version: Graylog 4.3.7+05bccc7

kingzacko1:
Previously, the actual format of the MS Teams Notification template didn't matter all that much. The code was parsing each line into a key value pair with a colon char delimiter so the `sections.facts` part of the [O365 connector card](https://learn.microsoft.com/en-us/microsoftteams/platform/task-modules-and-cards/cards/cards-reference#properties-of-the-office-365-connector-card), which relies on key-value pairings, could be used. If the value itself contained a colon then it would be cut off from that colon on. This broke timestamp fields and just didn't work all that well for things like backlog messages, which doesn't fit into a key-value pattern. 

These changes instead use the `text` field of the O365 connector card which allows for the template of the notification to include HTML or markdown more easily.

**This does introduce a breaking change for existing MSTeams Notifications using the old, default notification template**
image
Users will need to either fixup their markup beforehand so that it is actually used, or can update their notifications with the new default notification template that includes HTML. See Graylog2/graylog2-server#13738 for UPGRADING.md additions.

```
<table>
<tr><td><b>--- [Event Definition] ---</b></td></tr>
<tr><td><b>Title:</b></td><td>${event_definition_title}</td></tr>
<tr><td><b>Type:</b></td><td>${event_definition_type}</td></tr>

<tr><td><b>--- [Event] ---</b></td></tr>
<tr><td><b>Timestamp:</b></td><td>${event.timestamp}</td></tr>
<tr><td><b>Message:</b></td><td>${event.message}</td></tr>
<tr><td><b>Source:</b></td><td>${event.source}</td></tr>
<tr><td><b>Key:</b></td><td>${event.key}</td></tr>
<tr><td><b>Priority:</b></td><td>${event.priority}</td></tr>
<tr><td><b>Alert:</b></td><td>${event.alert}</td></tr>
<tr><td><b>Timestamp Processing:</b></td><td>${event.timestamp}</td></tr>
<tr><td><b>Timerange Start:</b></td><td>${event.timerange_start}</td></tr>
<tr><td><b>Timerange End:</b></td><td>${event.timerange_end}</td></tr>
<table>

<b>Event Fields:</b>
<table>
${foreach event.fields field}
<tr><td><b>${field.key}:</b></td><td>${field.value}</td></tr>
${end}
</table>

${if backlog}
<b>--- [Backlog] ---</b>
${foreach backlog message}
<pre>${message.timestamp}  ::  ${message.source}  ::  ${message.message}</pre>
${end}${end}
```

This new format should look almost identical to the previous one, except that fields with colons will actually render properly and so will the backlog. The `<pre>` tag may need to be tweaked if messages contained unescaped HTML characters.. but just removing them will work too. I'm also open to reworking that template and happily taking suggestions for improvements. Here is an example of a notification using the new code and template.

image

Also note the timestamps are now formatted differently, but I plan on cleaning that up in a separate PR addressing Graylog2/graylog2-server#13421

closes #1200
closes #1096

### DIFF
--- a/src/main/java/org/graylog/integrations/notifications/types/microsoftteams/TeamsEventNotification.java
+++ b/src/main/java/org/graylog/integrations/notifications/types/microsoftteams/TeamsEventNotification.java
@@ -16,7 +16,6 @@
  */
 package org.graylog.integrations.notifications.types.microsoftteams;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.floreysoft.jmte.Engine;
 import com.google.common.annotations.VisibleForTesting;
@@ -38,9 +37,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -114,7 +111,7 @@ public class TeamsEventNotification implements EventNotification {
     TeamsMessage createTeamsMessage(EventNotificationContext ctx, TeamsEventNotificationConfig config) throws PermanentEventNotificationException {
         String messageTitle = buildDefaultMessage(ctx);
         String description = buildMessageDescription(ctx);
-        JsonNode customMessage = null;
+        String customMessage = null;
         String template = config.customMessage();
         if (!isNullOrEmpty(template)) {
             customMessage = buildCustomMessage(ctx, config, template);
@@ -123,7 +120,7 @@ public class TeamsEventNotification implements EventNotification {
         TeamsMessage.Sections section = TeamsMessage.Sections.builder()
                 .activityImage(config.iconUrl())
                 .activitySubtitle(description)
-                .facts(customMessage)
+                .text(customMessage)
                 .build();
 
         return TeamsMessage.builder()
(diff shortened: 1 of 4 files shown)


ryan-carroll-graylog (approved):
Looks good and tests successfully

bernd (commented):
@kingzacko1 Can you add a section about the breaking change to the [UPGRADING.md](https://github.com/Graylog2/graylog2-server/blob/master/UPGRADING.md) file in the server repository? Thank you! :pray: 

Is there a way to do a non-breaking change that we can backport to fix the timestamp issue?

kingzacko1:
@bernd will do! Wanted to make sure that these changes looked good and were approved before updating that!

And yes, the [first commit](https://github.com/Graylog2/graylog-plugin-integrations/pull/1202/commits/774678170ae1bfa44c7ccdebe5319ef23acef26f) in this PR alone actually just did that alone. The backlog messages are still going to be a mess, but the timestamps should display properly. I can backport that commit by itself to 4.3 if we want to clean that one issue up there.

bernd:
> This looks good and everything worked well in testing.
> 
> Notifications with the old templates do look ugly but it appears all the info is still there at least. One thing that came to mind if we did want to deal with the backwards compatibility would be to update the `TeamsEventNotificationConfig` type to `teams-notification-v2` and handle the config types differently in `TeamsEventNotification.java`. I'm not sure how much extra complexity that would add and if the juice is worth the squeeze.

Thanks for testing @ryan-carroll-graylog! :pray: I think the main question regarding backward compatibility is if custom templates still work. When customers don't use the default template but their own.

ryan-carroll-graylog:
> > This looks good and everything worked well in testing.
> > Notifications with the old templates do look ugly but it appears all the info is still there at least. One thing that came to mind if we did want to deal with the backwards compatibility would be to update the `TeamsEventNotificationConfig` type to `teams-notification-v2` and handle the config types differently in `TeamsEventNotification.java`. I'm not sure how much extra complexity that would add and if the juice is worth the squeeze.
> 
> Thanks for testing @ryan-carroll-graylog! 🙏 I think the main question regarding backward compatibility is if custom templates still work. When customers don't use the default template but their own.

No problem! I think I was being ambiguous using "templates" when I meant  "Custom Message" templates.

Teams Notification Configs with user's old "Custom Message" template format do work after this change, but the formatting is off (e.g. multiple fields on the same line). All the data is still there it just won't be as pretty.

PR #1283, opened 2023-01-18
Update PaloAlto9xTemplates.java

giveen:
Addressed 
ERROR [PaloAltoTypeParser] Error parsing field application_response_time, auto is not a valid numeric value

## Notes for Reviewers

- [x] The commit history must be preserved - please use the rebase-merge or standard merge option instead of squash-merge
- [x] Sync up with the author before merging

### DIFF
--- a/src/main/java/org/graylog/integrations/inputs/paloalto9/PaloAlto9xTemplates.java
+++ b/src/main/java/org/graylog/integrations/inputs/paloalto9/PaloAlto9xTemplates.java
@@ -263,7 +263,7 @@ public class PaloAlto9xTemplates {
 
         fields.add(create(PaloAlto9xFields.PAN_LOG_PANORAMA, 35, STRING));
         fields.add(create(PaloAlto9xFields.PAN_SELECTION_TYPE, 36, STRING));
-        fields.add(create(ApplicationFields.APPLICATION_RESPONSE_TIME, 37, LONG));
+        fields.add(create(ApplicationFields.APPLICATION_RESPONSE_TIME, 37, STRING));
         fields.add(create(PaloAlto9xFields.PAN_GATEWAY_PRIORITY, 38, LONG));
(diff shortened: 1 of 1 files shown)


giveen:
@kingzacko1 can you take a look at this sometime soon?

giveen:
@ryan-carroll-graylog

kingzacko1:
@giveen do you have more information that you can provide as to why this change is necessary? The `application_response_time` field is one that should be coming in as a number and `auto` seems to rightly be being rejected because it is not a number.

kingzacko1:
@giveen From the [Palo Alto documentation](https://docs.paloaltonetworks.com/pan-os/10-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions/globalprotect-log-fields), it looks like that field is supposed to be in milliseconds. I'm curious where you are getting an `auto` from.

kingzacko1:
So after digging into this a bit, it looks like the issue is not the type of this field (it should indeed be `LONG`) but rather a missing column in the template. 

image

The `auto` value is actually for the `Selection Type` field, which should be a string. In that same templates file you've edited, we are missing the `High Res Timestamp` field. So where we have response time at the 37th index, it should actually be at 38. Once that change is made, the next issue is the `Priority` field at index 39. We have that marked as a `LONG` and according to docs it _should_ be a number 0-5, but based on some older [PAN articles](https://knowledgebase.paloaltonetworks.com/KCSArticleDetail?id=kA10g000000ClSsCAK) I found it seems like they may be putting the priority string value in that field instead of the integer value.

kingzacko1:
@giveen I have opened up an issue as well as a new PR with the necessary changelog and approvers added. I am going to close this PR, but you can follow along with progress on #1327 and #1328

giveen:
> 

THanks, I appreciate it.

PR #1320, opened 2023-02-27
Slack and MS Teams Notification Improvements

### LINKED ISSUE #780: Alert notifications for slack does not do @here
When adding/editing a Slack notification under the Alerts section, the option to notify @channel is there. I am missing a @here notification option as well, as that would only notify the users that are currently online and not everybody.

## What?
When adding/editing a Slack notification under the Alerts section, the option to notify @channel is there. I am missing a @here notification as well, as that would only notify the users that are currently online and not everybody.

## Why?
Currently, @channel notifies even users that are away/on mobile devices etc that really would not have to be disturbed. @here pings only the online users which makes more sense for certain notifications.

* Graylog Version: 4.0.5

### LINKED ISSUE #1172: Make the header in Slack notifications optional
# Description
Currently, the Slack notification code adds a hardcoded header to every message, which bloats the channel when many messages are received. The minimal header follows the following format:
```
Alert ${alertName} triggered:
Custom Message:
${customMessage}
```

## What
It should be possible to disable this hardcoded header through a checkbox in the UI, to minimize the height/length/size of the resulting Slack message. With the header disabled, the message should follow the following format instead:
```
${customMessage}
```
With the now optional header set to be enabled, the behavior should remain unchanged compared to now.

The relevant code is located [here](https://github.com/Graylog2/graylog-plugin-integrations/blob/master/src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java#L166) and a simple additional checkbox and if-check would be enough.

## Why
The header causes many short messages to become unreadable, because it takes up more space than the relevant part of the messages. This reduces productivity considerably.

### LINKED ISSUE #1318: Support timezone in Slack and MS Teams notifications
This is a follow-up to https://github.com/Graylog2/graylog2-server/issues/11418.

## What?

There should be a way to select the timezone for Slack and MS Teams notifications.

https://github.com/Graylog2/graylog2-server/pull/11982 already implemented this for email notifications.

## Why?

Currently the timestamps are UTC, which can be confusing for the users.

## Your Environment

* Graylog Version: 4.3.6

kingzacko1:
- Support time zones for Slack and Teams notifications (closes #1318)
  - Date times included in notifications should now be formatted with the configured time zone
  - Default time zone of UTC provided for existing notifications should ensure backwards compatibility
- Adds option to omit the title portion of a Slack notification to reduce clutter (closes #1172)
- Convert existing JSX files to TSX files - does not include a conversion to functional components.
  - Also removed a seemingly unused `graylog_url` config property
- Add support for notifying `@here` in Slack notifications instead of just `@channel` (closes #780)
  - Note that support is for EITHER channel OR here, not both

All of these changes should be backwards compatible and test successfully with existing MS Teams and Slack notifications

## Notes for Reviewers

### DIFF
--- a/src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
+++ b/src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
@@ -57,6 +57,7 @@ public abstract class SlackEventNotificationConfig implements EventNotificationC
     static final String INVALID_SLACK_URL_ERROR_MESSAGE = "Specified Webhook URL is not a valid Slack URL";
     static final String INVALID_DISCORD_URL_ERROR_MESSAGE = "Specified Webhook URL is not a valid Discord URL";
     static final String EMPTY_BODY_ERROR_MESSAGE = "If custom message is empty the title must be included";
+    static final String INVALID_NOTIFY_SETTINGS = "Can only notify either @channel or @here, not both.";
     static final String WEB_HOOK_URL = "https://hooks.slack.com/services/xxx/xxxx/xxxxxxxxxxxxxxxxxxx";
     static final String CHANNEL = "#general";
 
@@ -66,6 +67,7 @@ public abstract class SlackEventNotificationConfig implements EventNotificationC
     static final String FIELD_CUSTOM_MESSAGE = "custom_message";
     static final String FIELD_USER_NAME = "user_name";
     static final String FIELD_NOTIFY_CHANNEL = "notify_channel";
+    static final String FIELD_NOTIFY_HERE = "notify_here";
     static final String FIELD_LINK_NAMES = "link_names";
     static final String FIELD_ICON_URL = "icon_url";
     static final String FIELD_ICON_EMOJI = "icon_emoji";
@@ -115,6 +117,9 @@ public abstract class SlackEventNotificationConfig implements EventNotificationC
     @JsonProperty(FIELD_INCLUDE_TITLE)
     public abstract Boolean includeTitle();
 
+    @JsonProperty(FIELD_NOTIFY_HERE)
+    public abstract Boolean notifyHere();
+
     @Override
     @JsonIgnore
     public JobTriggerData toJobTriggerData(EventDto dto) {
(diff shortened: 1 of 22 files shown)


src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
```diff
@@ -104,6 +111,15 @@ public abstract class SlackEventNotificationConfig implements EventNotificationC
     @Nullable
     public abstract String iconEmoji();
 
+    @JsonProperty(FIELD_TIME_ZONE)
+    public abstract DateTimeZone timeZone();
```
ryan-carroll-graylog, line 115: @kingzacko1 I'm getting an error if I remove the time zone from the config and then execute a test notification (slack and teams). 

I'm also noticing that if I select `GMT-12` in the config I get `-12` in the resulting notification (and vice versa). Not sure if this is just me misunderstanding how that's supposed to work and you didn't touch any of that conversion code here.

```suggestion
    @Nullable
    public abstract DateTimeZone timeZone();
```
image

ryan-carroll-graylog (changes requested):
Other than the time zone comments this looks solid and tests successfully! Old notifications work fine and I really like the other improvements!

src/main/java/org/graylog/integrations/notifications/types/SlackEventNotificationConfig.java
kingzacko1: The idea was to require a time zone, but just default to UTC (as it has always been) if the field is never set. Let me remove the ability to clear the time zone. 

The `GMT-12` one may be a different issue, but I will check it out.

ryan-carroll-graylog: Ok makes sense to me. Just a heads up in case you're not already tracking, the UI says `(optional)` for that field so that should probably go too.

kingzacko1: @ryan-carroll-graylog and I don't understand it either... but look at [this list](https://www.ibm.com/docs/en/cloudpakw3700/2.3.0.0?topic=SS6PD2_2.3.0/doc/psapsys_restapi/time_zone_list.htm) of time zones and see that `Etc/GMT-12` is GMT+12 and `Etc/GMT+12` is GMT-12.... ¯\\\_(ツ)_/¯

ryan-carroll-graylog: Ah well there ya go. Figured it was something like that, with the alternative being that it been broken everywhere else we're doing similar and nobody noticed.

ryan-carroll-graylog (approved):
LGTM with the latest!

zeeklop (approved):
LGTM

PR #1335, opened 2023-04-14
fix(1334): use navigate in useEffect hook

zeeklop:
## Notes for Reviewers
Fixes a navigation issue that was causing a race condition where the HOC was expecting a modal view as a form. This input mounts a component that redirects the user to the stepper form view, unmounting the HOC, preventing the call to `this._openModal` on the `onClick` event for the `Lunc new input` button.

/nocl


danotorrey (approved):
LGTM and test successfully

